PR #13, opened 2020-07-31
Change CRDs domain to 'image.toolkit.fluxcd.io'

hiddeco:


config/crd/bases/toolkit.image.fluxcd.io_imagepolicies.yaml
```diff
@@ -0,0 +1,94 @@
+
```
squaremo, line 1: Where did this file come from?

config/crd/bases/toolkit.image.fluxcd.io_imagerepositories.yaml
```diff
@@ -0,0 +1,93 @@
+
```
squaremo, line 1: Where did this file come from?

config/crd/bases/toolkit.image.fluxcd.io_imagepolicies.yaml
hiddeco: Tutorial:

1) make a typo
2) generate the files
3) notice your typo and fix it
4) regenerate the files while not removing the old ones

PR #21, opened 2020-08-11
Update configuration to match the toolkit controller standard

stefanprodan:
Changes:
* adapt deployment config to match the toolkit kustomizations
* use Alpine as base image
* generate CRDs for `apiextensions.k8s.io/v1`
* configure JSON logging
* configure events dispatching


config/rbac/kustomization.yaml
```diff
@@ -1,12 +1,8 @@
...
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
-# Comment the following 4 lines if you want to disable
-# the auth proxy (https://github.com/brancz/kube-rbac-proxy)
-# which protects your /metrics endpoint.
-- auth_proxy_service.yaml
-- auth_proxy_role.yaml
-- auth_proxy_role_binding.yaml
-- auth_proxy_client_clusterrole.yaml
+namePrefix: image-reflector-
```
squaremo, line 8:
```suggestion
namePrefix: image-reflector-

```

squaremo: i.e., newline at the end of the file

config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
```diff
@@ -1,90 +1,89 @@
...
-              format: date-time
-              type: string
-          required:
-          - lastError
-          type: object
-      type: object
-  version: v1alpha1
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
```
squaremo, line 19: What happened here, is the reformat a result of changing the CRD version, or a difference in tooling?

Dockerfile
```diff
@@ -2,26 +2,30 @@
...
 COPY main.go main.go
 COPY api/ api/
 COPY controllers/ controllers/
 
-# Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+# build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o image-reflector-controller main.go
+
+FROM alpine:3.12
```
squaremo, line 21: alpine because useful for trouble-shooting?

main.go
```diff
@@ -94,3 +124,29 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// newLogger returns a logger configured for dev or production use.
```
squaremo, line 128: Seems like this is generic enough to go in pkg (in the future some time ..)

Dockerfile
stefanprodan: For CVE scanning as the distroless is not support by most scanners and we want something like Xray that can lookup the go modules and OS packages.

config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
stefanprodan: That's CRD v1 see https://kubernetes.io/blog/2019/06/20/crd-structural-schema/

main.go
stefanprodan: Maybe, I'm not convinced that this will not break helm-controller due to go-logr -> controller-runtime -> client-go <- helm

Dockerfile
stefanprodan: Another argument is consistency across the toolkit. Having the same base image with the same OS packages means that we could automate the base image update for all controllers.

stefanprodan: Ah I forgot about docker buildx and multi-arch builds.. if we want ARM and ARM64 then we have to use the alpine multi-arch base image or another image like Debian.

config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
squaremo: It's the same stuff as before, just shuffled around. So long as it won't happen every time the tooling is run, it doesn't really matter.

Dockerfile
squaremo: Didn't think of the CVE scanning thing, that's a good reason :+1:

main.go
squaremo: What is go-logr doing depending on controller-runtime?? Oh well

config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
stefanprodan: The format will not change unless you modify the API version in Makefile `CRD_OPTIONS ?= crd:crdVersions=v1`

PR #24, opened 2020-08-11
Implement Ready condition for image repos

stefanprodan:
Changes:
* add Ready condition to ImageRepository status
* set ImageRepository readiness
* return errors in the main reconciliation loop to record the failures as prom metrics


controllers/imagerepository_controller.go
```diff
@@ -113,6 +116,19 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 }
 
 func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo imagev1alpha1.ImageRepository, ref name.Reference) (imagev1alpha1.ImageRepository, error) {
+	// acquire lock
```
squaremo, line 119: I don't understand why this is necessary. The controller runtime ensured that Reconcile doesn't get run with the same resource concurrently, doesn't it?

stefanprodan: It doesn't, without the lock you can receive multiple updates events and they will run in parallel.

stefanprodan: To test this, increase the workers to 2 or more and run multiple kubectl apply, the repo will be scanned in parallel.

squaremo: Not processing the same resource concurrently is a documented feature of the workqueue: https://godoc.org/k8s.io/client-go/util/workqueue

stefanprodan: Hmm doesn't this apply to single worker? We decided to use the lock after getting parallel calls in helm operator.

stefanprodan: The workqueue should prevent concurrency across workers according to https://openkruise.io/en-us/blog/blog2.html

So as long as we don't branch to other go routines should be safe to remove the lock.

controllers/imagerepository_controller.go
```diff
@@ -72,55 +73,85 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+			corev1.ConditionFalse,
+			imagev1alpha1.ReconciliationFailedReason,
+			err.Error(),
+		)
+		if err := r.Status().Update(ctx, &status); err != nil {
+			return ctrl.Result{Requeue: true}, err
 		}
 		log.Error(err, "Unable to parse image name", "imageName", imageRepo.Spec.Image)
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, err
```
squaremo, line 86: I'm deliberately not returning an error -- the reconciliation _succeeded_, and the result was that the input is considered invalid.

More generally, not reaching a ready state does not mean that reconciliation failed. It may just be that some long-running process is still underway.

controllers/imagerepository_controller.go
```diff
@@ -72,55 +73,85 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-		tags, err := remote.ListWithContext(ctx, ref.Context()) // TODO: auth
-		if err != nil {
-			imageRepo.Status.LastError = err.Error()
-			if err = r.Status().Update(ctx, &imageRepo); err != nil {
-				return ctrl.Result{}, err
-			}
+
+		reconciledRepo, reconcileErr := r.scan(ctx, imageRepo, ref)
+		if err = r.Status().Update(ctx, &reconciledRepo); err != nil {
+			return ctrl.Result{Requeue: true}, err
```
squaremo, line 99: Doesn't returning an error guarantee that it'll be requeued? I guess it's helpful to be explicit about that.

controllers/imagerepository_controller.go
```diff
@@ -72,55 +73,85 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-		imageRepo.Status.LastScanTime = &metav1.Time{Time: now}
-		imageRepo.Status.LastScanResult.TagCount = len(tags)
-		imageRepo.Status.LastError = ""
-		// share the information in the database
-		r.Database.SetTags(canonicalName, tags)
-		log.Info("successful scan", "tag count", len(tags))
-		if err = r.Status().Update(ctx, &imageRepo); err != nil {
-			return ctrl.Result{}, err
+		if reconcileErr != nil {
+			return ctrl.Result{RequeueAfter: when}, reconcileErr
```
squaremo, line 103: I'm not clear on whether controller-runtime will ignore the RequeueAfter (because there's an error) or respect it. If it's the latter, a comment on why it's being requeued at the scheduled time, rather than letting it back off, might make sense.

controllers/imagerepository_controller.go
```diff
@@ -72,55 +73,85 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+				time.Now().Sub(now).String(),
+				when),
+			)
 		}
 	}
+
 	return ctrl.Result{RequeueAfter: when}, nil
 }
 
+func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo imagev1alpha1.ImageRepository, ref name.Reference) (imagev1alpha1.ImageRepository, error) {
```
squaremo, line 115: Good refactoring, thanks :top:

controllers/imagerepository_controller.go
stefanprodan: > I'm deliberately not returning an error -- the reconciliation succeeded, and the result was that the input is considered invalid.

If the spec is invalid we should send an error alert to notification-controller and record it as a failed reconciliation, same way we do for sources https://github.com/fluxcd/source-controller/blob/master/controllers/helmrepository_controller.go#L176

Are you suggesting to change the way validation errors are recorded and ignore them?

> More generally, not reaching a ready state does not mean that reconciliation failed. 

Ok but in this case it's a validation error. We could have a dedicated status reason like https://github.com/fluxcd/source-controller/blob/master/api/v1alpha1/condition_types.go#L62

> It may just be that some long-running process is still underway.

We signal this by setting the ready condition to unknown and the reason to progressing. In that case we would not return an error.

squaremo: Yes that's a fair point, invalid input is different to just failing to make progress. This is what validation hooks are for I presume, though a bit of a sledgehammer.

It's the owner of the resource that needs to see the problem, rather than the operator of the controller, so it feels like a condition is needed either way.

If the image string parsed OK, but wasn't accessible (either failure to authenticate, or didn't exist -- which may not be distinguishable cases), I'd use a condition to signal that the resource will not make progress without intervention. Arguably, a misshapen image string is closer to that situation than to a reconciliation failure. But it's a judgement call, I think.

controllers/imagerepository_controller.go
stefanprodan: I think we could return an error type from scan to signal rate limit errors and requeue right away, so that the runtime backoff with kick in. For other types of errors we could use the interval like we do for sources/helmreleases/kustomizations/etc.

squaremo: Yeah that's not a bad idea, to let the backoff kick in, if we get rate limited.
In the context of this PR -- is it useful to return both an error and `ctrl.Result{RequeueAfter: when}`?

stefanprodan: I've changed it to let the backoff kick in.

controllers/imagerepository_controller.go
stefanprodan: I've added an URL invalid reason as we do in source controller.

squaremo: My conclusion is: let's have a dedicated status reason for being unable to scan the repo. There's no point in requeuing it, because it needs to be changed before it'll work; and, it's not an unexpected failure of the controller, and it doesn't need to be logged; so, no reason to return an error from Reconcile.

stefanprodan: I've changed it to return no error.

controllers/imagerepository_controller.go
```diff
@@ -72,55 +73,85 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-		imageRepo.Status.LastScanTime = &metav1.Time{Time: now}
-		imageRepo.Status.LastScanResult.TagCount = len(tags)
-		imageRepo.Status.LastError = ""
-		// share the information in the database
-		r.Database.SetTags(canonicalName, tags)
-		log.Info("successful scan", "tag count", len(tags))
-		if err = r.Status().Update(ctx, &imageRepo); err != nil {
-			return ctrl.Result{}, err
+		if reconcileErr != nil {
+			return ctrl.Result{Requeue: true}, reconcileErr
```
squaremo, line 103: This will still show up as a reconciliation error, when it might be because e.g., the registry scan failed (i.e., a failure to make progress). I see that e.g., the source-controller does a similar thing -- perhaps it's worth having a (GH) discussion about it?

stefanprodan: In my mind a reconciliation error should bubble up no matter what the controller does: registry scan, git cloning, kubectl apply, helm install, etc. If we don't consider those as reconciliation errors then we should delete the monitoring stack as it becomes useless.

squaremo: https://github.com/fluxcd/toolkit/discussions/164

> If we don't consider those as reconciliation errors then we should delete the monitoring stack as it becomes useless

That's a bit over-dramatic. There's nothing stopping you recording your own metrics for failures to converge.

squaremo:
Now that I've tabled the discussion about [errors v no progress](https://github.com/fluxcd/toolkit/discussions/164) I'll get out of the way of this PR -- let's just do here what's done elsewhere, and return an error. If that discussion goes the way of making a distinct metric for ready states or whatever, we can change all the controllers to suit.

squaremo (approved):

controllers/imagerepository_controller.go
squaremo: (changed back, subsequently; see https://github.com/fluxcd/image-reflector-controller/pull/24#issuecomment-677647182)

controllers/imagerepository_controller.go
squaremo: See https://github.com/fluxcd/image-reflector-controller/pull/24#issuecomment-677647182

PR #27, opened 2020-08-18
Make the api package a dedicated module

hiddeco:


squaremo (approved):
OK cool, this will allow it to be imported without dragging in controller-runtime or other bits, that's the idea right?

PR #28, opened 2020-08-19
Fix bug in kustomization configuration.

bigkevmcd:
The kustomization.yaml was referring to base directories as resources,
which means that it couldn't be applied.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,9 +1,9 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: image-reflector-system
-resources:
+bases:
 - ../crd
 - ../rbac
 - ../manager
+resources:
 - namespace.yaml
-


bigkevmcd:
There is still an issue, in that the issue referred to either doesn't exist, or isn't public.

config/default/kustomization.yaml
```diff
@@ -1,9 +1,10 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: image-reflector-system
-resources:
+bases:
 - ../crd
 - ../rbac
 - ../manager
+resources:
 - namespace.yaml
```
hiddeco, line 10: While we are at it, can you remove this additional line?

hiddeco (approved):
LGTM, and in line with the other Toolkit components :1st_place_medal:

> There is still an issue, in that the image referred to (`fluxcd/image-reflector-controller`) either doesn't exist, or isn't public.

There has not been a release yet, the current configuration assumes an import of a locally build image on a (local) development cluster.

stefanprodan:
Kustomize `base` has been deprecated long time ago, please update your kustomize binary. We use `resources` only for the toolkit components https://github.com/fluxcd/toolkit/blob/master/manifests/install/kustomization.yaml

bigkevmcd:
> Kustomize `base` has been deprecated long time ago, please update your kustomize binary. We use `resources` only for the toolkit components https://github.com/fluxcd/toolkit/blob/master/manifests/install/kustomization.yaml

Ahhh...the subtleties of the kubectl version of kustomize (2.0.x) vs the external version (3.0.x).

Should've thought of that, thanks :-)

PR #29, opened 2020-08-25
Switch out the current semver parsing package.

bigkevmcd:
This replaces the current use of Masterminds/semver for blang/semver.

This is perhaps a more compliant package, but it does change the way that
prerelease versions are determined.

Previously 1.0.x did not match 1.1.0-beta but now, to get the same behaviour, use <1.1.0-0.

### DIFF
--- a/controllers/imagepolicy_controller.go
+++ b/controllers/imagepolicy_controller.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"strings"
 
-	semver "github.com/Masterminds/semver/v3"
+	semver "github.com/blang/semver/v4"
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


squaremo:
Sorry to hold this up -- I would like to see fluxcd/source-controller#127 resolved, so we can have consistency without also having broken semver ranges.

bigkevmcd:
@squaremo np, I saw that ticket opened, and I think resolving that is for the best...not entirely sure what the resolution is tho' but one way or another, I think we're agreed `1.1.0-pre` should not be in `1.0.x`

squaremo:
The resolution of fluxcd/source-controller#127 turns out to be .... port everything else over to Masterminds/semver :-)
(There's an issue already -- https://github.com/fluxcd/image-reflector-controller/issues/38 -- for using the version helper in pkg/version)

Issue #19, opened 2020-08-10
Suspend / Force scanning


squaremo:
The general pattern within GitOps Toolkit is that resources can be suspended (don't act on it), and those running on a schedule can be run on demand (act on it now). See, for example, https://toolkit.fluxcd.io/components/kustomize/kustomization/#reconciliation

This applies at least to ImageRepository. I don't think it applies to ImagePolicy, since it just reflects a calculation, rather than an ongoing process.

bigkevmcd:
For the second part (force-scan) I wonder if it wouldn't be useful to extract the [predicate](https://github.com/fluxcd/source-controller/blob/master/controllers/predicate.go) and annotation to be shared across services?

squaremo:
> For the second part (force-scan) I wonder if it wouldn't be useful to extract the [predicate](https://github.com/fluxcd/source-controller/blob/master/controllers/predicate.go) and annotation to be shared across services?

Yeah I think that's a good idea. There will probably still need to be code for deciding whether to do a scan, or if not, when to requeue. But it's a good habit to get into with the controllers, and will avoid some needless reconcile calls no doubt.

Side note: I initially didn't grok how scheduling via `RequeueAfter` would work with the filter. For the benefit of people as slow to catch on as I am, the predicate only pertains to the changes coming _into_ the queue, and the scheduling works by requeuing directly.

PR #30, opened 2020-08-26
Add support for suspending image scans.

bigkevmcd:
This adds support for suspending the scanning of an ImageRepository.

Addresses the first part of #19

### DIFF
--- a/api/v1alpha1/imagerepository_types.go
+++ b/api/v1alpha1/imagerepository_types.go
@@ -33,6 +33,11 @@ type ImageRepositorySpec struct {
 	// scans of the image repository.
 	// +optional
 	ScanInterval *metav1.Duration `json:"scanInterval,omitempty"`
+
+	// This flag tells the controller to suspend subsequent image scans.
+	// It does not apply to already started scans. Defaults to false.
+	// +optional
+	Suspend bool `json:"suspend,omitempty"`
 }
 
 type ScanResult struct {
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -77,6 +77,22 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 
 	log := r.Log.WithValues("controller", strings.ToLower(imagev1alpha1.ImageRepositoryKind), "request", req.NamespacedName)
 
+	if imageRepo.Spec.Suspend {
+		msg := "ImageRepository is suspended, skipping reconciliation"
+		status := imagev1alpha1.SetImageRepositoryReadiness(
+			imageRepo,
+			corev1.ConditionFalse,
+			imagev1alpha1.SuspendedReason,
+			msg,
+		)
+		if err := r.Status().Update(ctx, &status); err != nil {
+			log.Error(err, "unable to update status")
+			return ctrl.Result{Requeue: true}, err
+		}
+		log.Info(msg)
+		return ctrl.Result{}, nil
+	}
+
 	ref, err := name.ParseReference(imageRepo.Spec.Image)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


api/v1alpha1/imagerepository_types.go
```diff
@@ -33,6 +33,11 @@ type ImageRepositorySpec struct {
 	// scans of the image repository.
 	// +optional
 	ScanInterval *metav1.Duration `json:"scanInterval,omitempty"`
+
+	// This flag tells the controller to suspend subsequent kustomize executions,
+	// it does not apply to already started executions. Defaults to false.
+	// +optional
+	Suspend bool `json:"suspend,omitempty"`
```
squaremo, line 40: In general I'd be surprised to set a field to a meaningful value, then have that field omitted when reported back to me. Here, it's arguable whether `false` and missing are worth distinguishing, since the expected mode of use is to add the flag and _remove_ it, rather than toggling it between `true` and `false`. Since this is (I presume) following the convention used elsewhere in GOTK, I think that argument carries -- but I know future me would appreciate having a comment here about why the general norm is broken, so he doesn't have to think it through again.

controllers/imagerepository_controller.go
```diff
@@ -77,6 +77,22 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 
 	log := r.Log.WithValues("controller", strings.ToLower(imagev1alpha1.ImageRepositoryKind), "request", req.NamespacedName)
 
+	if imageRepo.Spec.Suspend {
+		msg := "ImageRepository is suspended, skipping reconciliation"
```
squaremo, line 81: Nit pick on the language: reconciliation is completed, not skipped; the thing being skipped is scanning of the image repository.

controllers/scan_test.go
```diff
@@ -101,6 +108,45 @@ var _ = Describe("ImageRepository controller", func() {
 		Expect(repoAfter.Status.CanonicalImageName).To(Equal(imgRepo))
 		Expect(repoAfter.Status.LastScanResult.TagCount).To(Equal(len(versions)))
 	})
+
+	Context("when the ImageRepository is suspended", func() {
+		It("does not process the image", func() {
```
squaremo, line 113: :100: for the test case!

squaremo (changes requested):
Thank you! There's a bit of copypasta worth tidying up, and I've done some nit-picking that you can address with code or discussion, at your discretion :-) (The request changes is just to fix the field description)

controllers/imagerepository_controller.go
bigkevmcd: I retrained the same wording as https://github.com/fluxcd/kustomize-controller/blob/master/controllers/kustomization_controller.go#L110 to avoid confusion and this is definitely arguable, it's not really reconciled (as folks would understand it) because the reconcile loop terminates early.

squaremo: > I retained the same wording as [the kustomize controller]

Hmm, yeah, I think it's wrong there -- I have taken it up [in a discussion](https://github.com/fluxcd/toolkit/discussions/164). Let's stick with consistency here for now, as you prefer.

api/v1alpha1/imagerepository_types.go
```diff
@@ -33,6 +33,11 @@ type ImageRepositorySpec struct {
 	// scans of the image repository.
 	// +optional
 	ScanInterval *metav1.Duration `json:"scanInterval,omitempty"`
+
+	// This flag tells the controller to suspend subsequent kustomize executions,
```
squaremo, line 37: Sorry, I should have put my comment about the copypasta here, since the schema is just generated from this type.

squaremo (approved):
Print it! Thanks Kevin 🍫

PR #32, opened 2020-09-16
FOSSA scan enabled

idvoretskyi:
Ref.: https://github.com/cncf/foundation/issues/109

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,27 @@
+name: FOSSA
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "^1.14.x"
+      - run: go version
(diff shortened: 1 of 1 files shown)


idvoretskyi:
cc @stefanprodan @dholbach

stefanprodan:
This repo is using main instead of master.

squaremo (commented):
Can the API key be a secret, or is it not an issue?

.github/workflows/fossa.yml
```diff
@@ -0,0 +1,27 @@
+name: FOSSA
+on:
+  push:
+    branches: [master]
```
squaremo, line 4:
```suggestion
    branches: [main]
```

.github/workflows/fossa.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [master]
```
squaremo, line 6:
```suggestion
    branches: [main]
```

stefanprodan:
The API key can't be a secret as this is intended to run on PRs. All CNCF projects uses it like this so I don't think it's an issue.

PR #33, opened 2020-10-08
Add release workflow

squaremo:
This is copied in whole from source-controller, and adapted by replacing `source-controller` with `image-reflector-controller` (only).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup Kustomize
-        uses: fluxcd/pkg/actions/kustomize@master
+        uses: fluxcd/pkg/actions/kustomize@main
       - name: Prepare
         id: prep
         run: |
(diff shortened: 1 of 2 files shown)


squaremo:
(The only way to test this is to tag the repo)

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+        uses: actions/create-release@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          draft: false
+          prerelease: true
+          body: |
+            [CHANGELOG](https://github.com/fluxcd/image-reflector-controller/blob/master/CHANGELOG.md)
```
hiddeco, line 105:
```suggestion
            [CHANGELOG](https://github.com/fluxcd/image-reflector-controller/blob/main/CHANGELOG.md)
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+    tags:
+      - 'v*'
+
+jobs:
+  build-push:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Kustomize
+        uses: fluxcd/pkg/actions/kustomize@master
```
hiddeco, line 13:
```suggestion
        uses: fluxcd/pkg/actions/kustomize@main
```

.github/workflows/release.yml
squaremo: Good spot

hiddeco (approved):

PR #34, opened 2020-10-19
Respect the reconcileAt annotation

### LINKED ISSUE #19: Suspend / Force scanning

squaremo:
The GOTK controllers all use an annotation to trigger syncing outside of the schedule (e.g., when a webhook is indicates a new revision). This commit makes the ImageRepository controller respect the annotation, by checking it when examining an image repo object.

This works a little different to the other controllers, which tend to assume that any incoming event means a sync is due. The image reflector controller goes further in checking whether it's time to scan according to the annotation or the prescribed schedule.

Fixes #19 (in combination with #30).


hiddeco:
> This works a little different to the other controllers, which tend to assume that any incoming event means a sync is due. The image reflector controller does a bit more to check whether it's keeping to the prescribed schedule.

What is the idea behind this? As it defeats the purpose of why we originally invented the annotation (that being: requesting a resource to be reconciled, off-schedule).

squaremo:
> > This works a little different to the other controllers, which tend to assume that any incoming event means a sync is due. The image reflector controller does a bit more to check whether it's keeping to the prescribed schedule.
> 
> What is the idea behind this? As it defeats the purpose of why we originally invented the annotation (that being: requesting a resource to be reconciled, off-schedule).

I meant in the absence of the annotation. I'll restate it.

api/v1alpha1/imagerepository_types.go
```diff
@@ -62,6 +62,11 @@ type ImageRepositoryStatus struct {
 	// LastScanResult contains the number of fetched tags.
 	// +optional
 	LastScanResult ScanResult `json:"lastScanResult,omitempty"`
+
+	// LastHandledScanRequest reocrds the value of the annotation
```
bigkevmcd, line 66:
```suggestion
	// LastHandledScanRequest records the value of the annotation
```

This will need a rebuild of the CRD (because the comment in the CRD is derived from this).

bigkevmcd: But, I wonder if this shouldn't be named `LastHandledReconcileAt` https://github.com/fluxcd/kustomize-controller/blob/master/controllers/kustomization_controller.go#L264

bigkevmcd (commented):
Looks mostly good to me, I'd like to see standardisation of naming across the various tools, and the other is a typo :-)

api/v1alpha1/imagerepository_types.go
squaremo: > I wonder if this shouldn't be named `LastHandledReconcileAt`

Mm, yeah probably. I went with my initial instinct, to name it something meaningful in its context, given the [discussion over the annotation](https://github.com/fluxcd/toolkit/discussions/363). But: it is not especially visible, so better the name be consistent.

bigkevmcd: I read the discussion about naming, and yes, `reconcileAt` does feel a bit weird (it's like scheduling, but not), but I think consistency should win out (even if that means changing _other things_)

hiddeco (approved):
Looks :ok_hand: to me, thanks for making the changes to the other controllers too :100:

PR #35, opened 2020-10-21
Use credentials from secret

squaremo:
This lets you create a secret with registry credentials, then refer to it in the `ImageRepository` field `secretRef` so it can be used to scan image repositories that require authentication.


squaremo:
Among other things, this is waiting for me to write a test for using the authentication.

stefanprodan (approved):
LGTM

PR #39, opened 2020-10-30
Update k8s to 1.19

relu:
Note k8s 1.19 brings in klog v2, hence the update on the logr and respectively zapr (indirect, just for compatibility reasons) packages.


squaremo (approved):
LGTM, thank you @relu

PR #41, opened 2020-11-03
Adopt k8s standardized conditions

relu:
Use the newly introduced standardized Condition type kubernetes/enhancements#1624

Relates to fluxcd/flux2#225


squaremo (approved):
Happy with this ✅ , thank you @relu.

PR #45, opened 2020-11-17
Switch to new Condition helper funcs

relu:
Since we're now relying on the standardized metav1.Condition type
instead of our own implementation we should also make use of the helper
funcs provided by pkg/apis/meta and k8s api/meta.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -3,7 +3,7 @@ module github.com/fluxcd/image-reflector-controller/api
 go 1.15
 
 require (
-	github.com/fluxcd/pkg/apis/meta v0.3.0
+	github.com/fluxcd/pkg/apis/meta v0.4.0
 	k8s.io/api v0.19.3
 	k8s.io/apimachinery v0.19.3
 	sigs.k8s.io/controller-runtime v0.6.3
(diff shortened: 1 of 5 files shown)


relu:
Strange, I'm not able to reproduce the test failure. Ran the tests locally a few times with the same seed `-ginkgo.seed=1605887147` and they pass just fine.

squaremo:
Let me try here ..

squaremo:
Ah OK. `fluxcd/pkg/apis/meta.SetResourceCondition(...)` calls k8s.io/.../metav1.SetStatusCondition(...), which will only update the last transition time if the status has changed (and quite rightly -- there's no transition if it didn't change). Scanning the tags again doesn't transition _out_ of ready.

The test needs to check something that does change when the image repo is scanned again. There was a LastScanTime, but it was removed in d966c16a in favour of conditions. The scan result is the remaining possibility -- e.g., add a tag to the image repo, and check that it picks it up.

relu:
Got it, now it makes sense. I don't really get why it doesn't fail on my machine though, but I can definitely come up with a change based on your proposed solution.

I suppose `GetLastTransitionTime` has no real use now, so it can probably be removed. In the `shouldScan` method from the controller, the conditional return can be replaced by checking if the ready condition is not true.

relu:
Superseded by https://github.com/fluxcd/image-reflector-controller/pull/56

PR #53, opened 2020-11-19
Port the parameterised release workflow here

squaremo:
.. from image-automation-controller. It is exactly the same except for the name of the controller, used in various capacities.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,9 @@ on:
     tags:
       - 'v*'
 
+env:
+  CONTROLLER: image-reflector-controller
+
 jobs:
   build-push:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -49,8 +52,8 @@ jobs:
           file: ./Dockerfile
           platforms: linux/amd64
           tags: |
-            ghcr.io/fluxcd/image-reflector-controller:${{ steps.prep.outputs.VERSION }}
-            docker.io/fluxcd/image-reflector-controller:${{ steps.prep.outputs.VERSION }}
+            ghcr.io/fluxcd/${{ env.CONTROLLER }}:${{ steps.prep.outputs.VERSION }}
```
stefanprodan, line 55: I think you can use `${{ github.event.repository.name }}` instead of hardcoding the controller name.

squaremo: I'll try it!

stefanprodan:
I can’t comment on this line FROM golang:1.15 as builder but it should be golang-alpine

squaremo:
> I can’t comment on this line FROM golang:1.15 as builder but it should be FROM golang:1.15-alpine as builder

How would you expect this to fail -- failed `make docker-build`, or unusable image?

stefanprodan (approved):
LGTM 💯

.github/workflows/release.yml
squaremo: (yes it works!)

PR #56, opened 2020-11-23
Api refactoring

relu:
This is an attempt to bring the api and controller logic closer to what
the other controller components already have set as patterns.

1. Adopt the k8s standard Condition type.
2. Rename `ScanInterval` to `Interval` to be consistent with the
   `Interval` attribute other Spec types have defined, translating to
   reconciliation interval. This attribute is now required.
3. Add `ScanTime` attribute to the `ScanResult` type, enabling keeping
   track of the last successful scan execution. Use this value for scan
   frequency throttling.
4. Add optional `Timeout` attribute to allow custom scan timeout
   handling. The default value is equal to that of the `Interval` attr.

What is to be debated:

I proposed a different means of handling the reconciliation interval, while previously this was an optional value, it is not required. I think setting the interval should be a conscious decision made by the user considering the expectation of scan frequency and potential throttling imposed by the registry service provider. With this approach, we could recommend a default value while not imposing any restrictions aside from that of it being more than or equal to `1s`, the rest of the logic remains in place.
A potential future design proposed by @squaremo, is to **not** enforce a mandatory value and rather scan as often as possible if no value is provided, while also providing an attribute that would set an upper bound for throttling purposes (*at most* frequent).


stefanprodan:
> Rename ScanInterval to Interval to be consistent with the
Interval attribute other Spec types have defined, translating to
reconciliation interval. This attribute is now required.

Nice 👍

controllers/imagerepository_controller.go
```diff
@@ -192,30 +198,29 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo imagev1a
...
 	// never scanned; do it now
-	lastTransitionTime := imagev1alpha1.GetLastTransitionTime(repo)
-	if lastTransitionTime == nil {
+	lastScanTime := repo.Status.LastScanResult.ScanTime
+	if lastScanTime == nil {
 		return true, scanInterval
 	}
 
-	// Is the controller seeing this because the reconcileAt
+	// Is the controller seeing this because the reconcileAt return nil
```
squaremo, line 223: Copypasta

controllers/imagerepository_controller.go
```diff
@@ -70,6 +65,7 @@ type ImageRepositoryReconciler struct {
 
 func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
+	reconcileStart := time.Now()
```
squaremo, line 68: Do you need this _and_ `now` below?

squaremo (changes requested):
Looks good -- but there's a couple of things to fix up:
 - copypasta in a comment (see comment)
 - the commit message for the third commit doesn't match what that commit does (it doesn't introduce the standard Condition type, which is done earlier)

controllers/imagerepository_controller.go
relu: Good point! It's probably just a few milliseconds apart, but I think this is the right value to use regardless of that.

relu:
> the commit message for the third commit doesn't match what that commit does (it doesn't introduce the standard Condition type, which is done earlier)

Too much history rewriting :)
I squashed that one into the latest commit now, it should be fine. Thanks.

squaremo (approved):
All makes sense; thanks @relu 🥬

Issue #55, opened 2020-11-23
Add GitOps toolkit metrics


squaremo:
Both the GOTK conventional metrics, where they make sense, and any particular to image-reflector.

Issue #50, opened 2020-11-18
Send K8s events in gotk convention


squaremo:
Need to look for examples, first. Will likely need some design.

Issue #49, opened 2020-11-18
Use GOTK log setup and structured logging


squaremo:
Use the logger setup from https://github.com/fluxcd/pkg/tree/main/runtime/logger and review use of logging vs dev guide / examples elsewhere.

Issue #46, opened 2020-11-18
Retool in the GOTK style


squaremo:
I.e., do the same for this controller as https://github.com/fluxcd/image-automation-controller/pull/28 did.

PR #57, opened 2020-11-24
Toolkit compatibility

relu:
The set of changes touches on the following issues:
- Retool in the GOTK style #46
- liveness and readiness probes #48
- Use GOTK log setup and structured logging #49
- Add GitOps toolkit metrics #55
- Send K8s events in gotk convention #50
- Use pkg/version to parse versions #38


config/manager/deployment.yaml
```diff
@@ -25,14 +25,21 @@ spec:
...
-          httpGet:
-            port: http-prom
-            path: /metrics
+          - containerPort: 9440
+            name: healthz
+            protocol: TCP
         args:
-          - --enable-leader-election
-          - --log-level=debug
+          - --log-level=info
```
stefanprodan, line 32: The required args are:
```
          - --watch-all-namespaces
          - --log-level=info
          - --log-json
          - --enable-leader-election
```

relu: Yup, I have that in my stash, will commit.

controllers/imagepolicy_controller.go
```diff
@@ -89,19 +104,23 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 			return ctrl.Result{}, err
 		}
-		if latest != "" {
-			pol.Status.LatestImage = repo.Spec.Image + ":" + latest
-			err = r.Status().Update(ctx, &pol)
-		}
-		return ctrl.Result{}, err
-	default:
-		// no recognised policy, do nothing
+		r.event(pol, events.EventSeverityInfo, fmt.Sprintf("Latest image tag resolved to: %s", latest))
```
stefanprodan, line 123: Can we add the full image URI to the event please? Maybe something like:
```
fmt.Sprintf("Latest image tag for '%s' resolved to: %s", repo.Spec.Image, latest)
```

Dockerfile
```diff
@@ -19,14 +19,20 @@ COPY main.go main.go
...
 
+LABEL org.opencontainers.image.source="https://github.com/fluxcd/image-reflector-controller"
+
 RUN apk add --no-cache ca-certificates tini
 
 COPY --from=builder /workspace/image-reflector-controller /usr/local/bin/
 
+# Create minimal nsswitch.conf file to prioritize the usage of /etc/hosts over DNS queries.
+# https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-354316460
+RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
```
squaremo, line 34: Would this have a bit better caching behaviour if it were above the COPY directive?

controllers/imagepolicy_controller.go
```diff
@@ -89,19 +104,23 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 
 	policy := pol.Spec.Policy
 
-	switch {
-	case policy.SemVer != nil:
-		latest, err := r.calculateLatestImageSemver(&policy, repo.Status.CanonicalImageName)
-		if err != nil {
+	if policy.SemVer == nil {
```
squaremo, line 107: The reason this is a `switch` is that there are likely to be other kinds of policy, imminently. When that's the case, it'll just be rewritten back again -- I'd prefer to just keep the current form and avoid that.

controllers/imagepolicy_controller.go
```diff
@@ -164,3 +183,28 @@ func (r *ImagePolicyReconciler) imagePoliciesForRepository(obj handler.MapObject
 	}
 	return reqs
 }
+
+// event emits a Kubernetes event and forwards the event to notification controller if configured
+func (r *ImagePolicyReconciler) event(policy imagev1alpha1.ImagePolicy, severity, msg string) {
+	if r.EventRecorder != nil {
+		r.EventRecorder.Eventf(&policy, "Normal", severity, msg)
```
squaremo, line 190: Why use `Eventf`?

controllers/imagepolicy_controller.go
```diff
@@ -164,3 +183,28 @@ func (r *ImagePolicyReconciler) imagePoliciesForRepository(obj handler.MapObject
...
+	}
+	if r.ExternalEventRecorder != nil {
+		objRef, err := reference.GetReference(r.Scheme, &policy)
+		if err != nil {
```
squaremo, line 194: Oh for an error monad. Can you collapse these two identical error-handling clauses into one? I use

```
err := operation1()
if err == nil {
  err = operation2()
}
if err != nil {
  // handle error
}
```

but some people find that distasteful (and being honest, I have tripped myself up with it in the past :-/)

controllers/imagerepository_controller.go
```diff
@@ -122,12 +137,17 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-		if err = r.Status().Update(ctx, &imageRepo); err != nil {
+		if err := r.Status().Update(ctx, &imageRepo); err != nil {
 			return ctrl.Result{Requeue: true}, err
 		}
 		if reconcileErr != nil {
+			r.event(imageRepo, events.EventSeverityError, reconcileErr.Error())
 			return ctrl.Result{Requeue: true}, reconcileErr
 		}
+		// emit successful scan event
+		if rc := apimeta.FindStatusCondition(imageRepo.Status.Conditions, meta.ReconciliationSucceededReason); rc != nil {
```
squaremo, line 148: If this isn't straight-forwardly contingent on getting a `nil` error return, I would consider returning a bool saying whether a scan was done or not, to avoid this indirect way of finding out.

controllers/imagerepository_controller.go
```diff
@@ -277,3 +301,51 @@ func authFromSecret(secret corev1.Secret, registry string) (authn.Authenticator,
...
+			r.Log.WithValues(
+				"request",
+				fmt.Sprintf("%s/%s", repo.GetNamespace(), repo.GetName()),
+			).Error(err, "unable to send event")
+			return
+		}
+	}
+}
+
+func (r *ImageRepositoryReconciler) recordReadiness(repo imagev1alpha1.ImageRepository) {
```
squaremo, line 330: I didn't know what this method was for until I read through its code -- can we name it `recordReadinessMetric` instead?

controllers/imagerepository_controller.go
```diff
@@ -91,10 +95,20 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
 		return ctrl.Result{}, nil
 	}
 
+	// record reconciliation duration
+	if r.MetricsRecorder != nil {
+		objRef, err := reference.GetReference(r.Scheme, &imageRepo)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		defer r.MetricsRecorder.RecordDuration(*objRef, reconcileStart)
```
squaremo, line 109: Could the `recordReadiness` also be deferred -- it appears to be just before every exit.

squaremo (commented):
Nothing amiss here, but I had some suggestions, which you can take or leave :-)

Dockerfile
relu: Good point! I just copied it over as seen in other components' Dockerfiles. I will adjust, can probably be the first directive considering it's static nature.

controllers/imagepolicy_controller.go
relu: I understand, thanks!

controllers/imagepolicy_controller.go
relu: As you can imagine, I did a lot of copy-pasting :smile:

controllers/imagerepository_controller.go
relu: By the time we reach this, a scan was most definitely completed.

`rc` is not of error type, it's `*metav1.Condition`, can be `nil` if no condition was found (should never happen). This is used just to get the condition `Message` used as the event message. Does that make any sense?

controllers/imagerepository_controller.go
relu: Makes sense to me. I suppose this is something worth applying to other controller components that define it.

controllers/imagerepository_controller.go
squaremo: OIC, I didn't notice the message was used. OK, that's fair enough!

controllers/imagerepository_controller.go
```diff
@@ -79,6 +83,9 @@ func (r *ImageRepositoryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 
 	log := r.Log.WithValues("controller", strings.ToLower(imagev1alpha1.ImageRepositoryKind), "request", req.NamespacedName)
 
+	// record rediness metric
+	defer r.recordReadinessMetric(imageRepo)
```
squaremo, line 87: This passes imageRepo by value, including the value of the status and the value of the conditions slice. This means the conditions examined by recordReadinessMetrics will be whatever is the case at this line, rather than whatever has been calculated by the end of `Reconcile`. To fix this, pass a pointer, or if you don't want to depart from the convention set by other controllers, use a thunk `func() { r.recordReadinessMetric(imageRepo) }` so the evaluation is delayed.

relu: Great observation, I will change it to pass it by reference and I think this pattern should be followed by other controllers too, it does seem to be a common thing that we always exit the reconciliation loop immediately after this call is made.

squaremo (approved):
This crosses out lots of TODO items ☑️ ☑️ ☑️ , amazing. Thanks @relu 🍍

controllers/imagerepository_controller.go
squaremo: There's a fair bit of stuff (methods like `event()` and `recordReadinessMetric()`, but maybe control flow too) that could be factored out into a GOTK controller shell. I may open a discussion about it, but if someone beat me to that I would not be upset :-)

relu: Yes, I've noticed this too while jumping from one repo to the other, quite a lot of common code block candidates eligible for extraction into `pkg` maybe.

Issue #31, opened 2020-09-04
Lexicographic sort policy


bigkevmcd:
There are some tag conventions that lend themselves naturally to lexicographic sorting, for example, nightly builds with a timestamp style date, or [CalVer](https://calver.org/) style naming.

It'd be nice to support this too, essentially the sorted list of names, and take the last one.

relu:
Closing as it was implemented in #58

PR #58, opened 2020-11-24
Implement Alphabetical order policy

relu:
This implementation allows one to set a `Prefix` to filter a list of tags and/or use `Reverse` alphabetical order through the respective policy type attributes.

Relates to #31


controllers/imagepolicy_controller.go
```diff
@@ -89,22 +91,26 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
-		latest, err := r.calculateLatestImageSemver(&policy, repo.Status.CanonicalImageName)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		if latest != "" {
-			pol.Status.LatestImage = repo.Spec.Image + ":" + latest
-			err = r.Status().Update(ctx, &pol)
-		}
-		return ctrl.Result{}, err
+		latest, err = r.calculateLatestImageSemver(&policy, repo.Status.CanonicalImageName)
```
bigkevmcd, line 98: I wonder if we couldn't break these out into implementations of a _Policy_ interface, with a simple map translation here, which would make it easier to test the policies "standalone" ?

api/v1alpha1/imagepolicy_types.go
```diff
@@ -53,6 +56,18 @@ type SemVerPolicy struct {
...
+// AlphabeticalPolicy specifices a alphabetical ordering policy.
+type AlphabeticalPolicy struct {
+	// Prefix enables filtering the list of tags by given prefix.
+	// +optional
+	Prefix string `json:"prefix,omitempty"`
+	// Reversed enables using reversed alphabetical order.
+	// If true the order would be: C B A (A would be chosen)
+	// If false the order would be: A B C (C would be chosen)
+	// +optional
+	Reversed *bool `json:"reverse,omitempty"`
```
squaremo, line 68: An alternative, more familiar to some people, is using `order: ASC|DESC` (as in "ascending" or "descending"). The intuitive default would be ascending, which is the same as `reverse: false`, the default here and the most useful for e.g., calendar-dated tags.

api/v1alpha1/imagepolicy_types.go
```diff
@@ -53,6 +56,18 @@ type SemVerPolicy struct {
 	Range string `json:"range"`
 }
 
+// AlphabeticalPolicy specifices a alphabetical ordering policy.
+type AlphabeticalPolicy struct {
+	// Prefix enables filtering the list of tags by given prefix.
+	// +optional
+	Prefix string `json:"prefix,omitempty"`
```
squaremo, line 63: Is this a pattern or a literal? Does the matching prefix get removed before ordering? (Worth putting these considerations in the docs!)

controllers/imagepolicy_controller.go
squaremo: What does "map translation" mean here -- dispatch via a map of policy name -> implementation?

squaremo (commented):
Just some comments, while this is a draft ..

controllers/imagepolicy_controller.go
relu: I think that makes sense. I'm going to try to come up with something.

api/v1alpha1/imagepolicy_types.go
relu: My thinking was based more on the meaning of the adjective, which implies the default ascending order according to the alphabet, so *reversed* came to mind as the right modifier word. I would be inclined to change it if you think *order* makes more sense.

api/v1alpha1/imagepolicy_types.go
relu: Right now it's just a literal, using `strings.HasPrefix`. We can discuss the possibility to use a pattern, though I wasn't able to think of any use case for that, however, some people might find support for that useful. What do you think?

As a side note, zooming out a bit, I think it might make sense to provide the ability to filter out tags at the level of the imagerepository resource as well, in order to prevent polluting the tags database with any unused entries.

controllers/imagepolicy_controller.go
bigkevmcd: @squaremo yes

api/v1alpha1/imagepolicy_types.go
bigkevmcd: I think you need to clarify the intent of the prefix, i.e. is it matching, or excluding.

relu: I would think users might want to limit the tags the policy would apply to only to a subset of tags e.g `rel-*` or `nightly-*`. The intention was to use `Prefix` for matching this subset, excluding other irrelevant tags.

Would renaming this attribute to `MatchTagPrefix` make more sense?
What about supporting multiple values?
Or maybe adding support for regular expressions?

relu: Made some changes, however, some of the questions I mentioned above are still valid.

controllers/imagepolicy_controller.go
relu: I've moved the logic to an internal `policy` package and defined there the policy types and a factory to resolve the type based on the `ImagePolicyChoice` object. I think this is a lot better than the switch and allows us to extend further if needed without involving any controller logic changes.

internal/policy/factory.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+
+	imagev1 "github.com/fluxcd/image-reflector-controller/api/v1alpha1"
+)
+
+// Policy constructs a new policy object based on
+func Policy(choice imagev1.ImagePolicyChoice) (Interface, error) {
+	var p Interface
+	var err error
+	switch {
```
squaremo, line 30: Yep this works well :+1:

internal/policy/policy.go
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package policy
+
+type Interface interface {
```
squaremo, line 19: I'd name this something that describes what it does. (The interface in go's sort package defines the single requirement for `sort`, which doesn't typically need to be named outside of its own package -- so it can be forgiven for being named `sort.Interface`.)

squaremo (commented):
I had some minor nits to pick, so I've review-commented. We can also discuss the questions left to resolve, in the main comment thread.

api/v1alpha1/imagepolicy_types.go
```diff
@@ -53,6 +56,21 @@ type SemVerPolicy struct {
...
+	// character strings reprezenting the alphabet, ascending order will return
+	// the value 'z', while descending order will return the value 'a'.
+	// +kubebuilder:validation:Enum=asc;desc
+	// +optional
+	Order string `json:"order,omitempty"`
+	// MatchTagPrefix enables filtering for only a subset of tags based on their
+	// literal prefix. If not provided, all the tags from the repository will be
+	// ordered and compared.
+	// +optional
+	MatchTagPrefix string `json:"matchTagPrefix,omitempty"`
```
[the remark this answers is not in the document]
squaremo: (posting this in the main comment thread)

squaremo:
Open questions:

 - how to leave room for other ways of filtering later -- e.g., would it ever make sense to have prefix matching and regular expressions, and if not, how to prevent those both being supplied;
 - whether this is useful for all selection policies -- maybe you tag your images dev-<semver>;
 - if you might need to remove the prefix before considering the remainder for sorting, as in the dev-<semver> example.

controllers/imagepolicy_controller.go
```diff
@@ -103,12 +102,12 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 		return ctrl.Result{}, nil
 	}
 
-	policy := pol.Spec.Policy
+	resolver, err := policy.Policy(pol.Spec.Policy)
```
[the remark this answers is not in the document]
relu: Oh man, I'm bad at naming things :man_facepalming:

internal/policy/policy.go
relu: Fine, then I'll name it `Policer` (I had to check if that word exists and was pleasantly surprised it does).

relu:
> how to leave room for other ways of filtering later -- e.g., would it ever make sense to have prefix matching and regular expressions, and if not, how to prevent those both being supplied;

I would argue for policy-scoped filtering rather than per policy type implementation, for me this makes more sense because of my previous experience working with [ECR Lifecycle Policies](https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lp_tag_prefix_list) that also enable applying policies based on a prefix-based list of strings.

> whether this is useful for all selection policies -- maybe you tag your images dev-;

I think this can be useful even for semver-like tags that maybe have an incompatible prefix e.g. `ver1.2.3` instead of the supported format `v1.2.3`. This would enable folks to make use of semver policying w/o having to change the format of their tags, or in the situation that would not even be possible due to the repository being outside of their control.

> if you might need to remove the prefix before considering the remainder for sorting, as in the dev- example.

An inclusive/exclusive comparison mode toggle would be useful considering the case highlighted above.

relu:
An extra thought on the policy level prefix matching feature. I think it would be best if we implement this after we integrate a "real" database into the controller so that we can use its capabilities to filter out the tags dataset.

I'm inclined to remove the prefix matching implementation done in the scope of the Alphabetical ordering policy model type.

api/v1alpha1/imagepolicy_types.go
```diff
@@ -34,6 +34,11 @@ type ImagePolicySpec struct {
 	// selecting the most recent image
```
relu, line 34: Here's what I had in mind re: https://github.com/fluxcd/image-reflector-controller/issues/62#issuecomment-737546210

squaremo:
> I'm inclined to remove the prefix matching implementation done in the scope of the Alphabetical ordering policy model type.

As in, not include the prefix matching here, and just have alphabetical ordering? If so I think that's a good idea, until we resolve those discussions above.

relu:
> As in, not include the prefix matching here, and just have alphabetical ordering? If so I think that's a good idea, until we resolve those discussions above.

You can check the currently proposed design, I've moved the prefix filtering into the policy spec itself rather than at the level of the policy type implementation.

api/v1alpha1/imagepolicy_types.go
```diff
@@ -55,6 +63,33 @@ type SemVerPolicy struct {
...
+	// alphabet as tags, ascending order would select Z, and descending order
+	// would select A.
+	// +kubebuilder:default:="asc"
+	// +kubebuilder:validation:Enum=asc;desc
+	// +optional
+	Order string `json:"order,omitempty"`
+}
+
+// TagPrefixMatcher enables filtering tags based on a set of defined rules
+type TagPrefixMatcher struct {
```
squaremo, line 78: I like how this is expressed. At risk of exploding the scope, is it ever useful to match by suffix? Or using a glob? (I would count regexp out, per xkcd).

relu: Considering there are images that have suffixes such as `-alpine`, `-slim`, `-rc` etc. it might be a valid use case. I'm not sure what's the best approach around dealing with both of these use cases, I kind of like more the idea of globbing patterns vs prefix/suffix matching, would probably make sense.

relu: Using something like [this](https://github.com/gobwas/glob#example) maybe?

The only downside moving to glob patterns would probably be that we wouldn't be able to support the current pre-sort "trimming" functionality.

I'm trying to figure out, as a user, what would bring in the most value at the moment.

squaremo: Hmm yeah glob libraries don't tend to implement captures.

Meta-suggestion: shall we move this discussion to an issue or new PR, and reduce this PR to just the alphabetical ordering and related refactoring? That would still be pretty helpful, e.g., for calendar-versioned images.

relu: I agree, will move out the prefix filtering to a separate branch and open a PR once this one gets merged.

controllers/imagepolicy_controller.go
```diff
@@ -174,7 +175,7 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	}
 	r.event(pol, events.EventSeverityInfo, msg)
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
```
squaremo, line 178: Not sure of the intent of this change, though it doesn't change the meaning of the code as it stands.

squaremo (approved):
Neat! Thanks @relu 🍍

controllers/imagepolicy_controller.go
relu: Likely a relic from previous changes made that are no longer part of this PR. Should probably be changed back to `nil`.

api/v1alpha1/imagepolicy_types.go
nomeelnoj: @relu i see your branch for this, but do not see an open PR.  Dont want to step on any toes, but can we open this PR? Our currently workflow prevents us from moving to the Helm Controller until we have this functionality, but the helm operator is breaking due to the deprecation of the old stable repo, so we are stuck currently: https://github.com/fluxcd/image-reflector-controller/tree/tag-prefix-matching

PR #59, opened 2020-11-26
Use pkg/controller helpers

squaremo:
This is an experimental, ongoing, non-invasive rewrite of the image-reflector controller to use the controller helpers from https://github.com/fluxcd/pkg/pull/56.


squaremo:
Closing as abandoned; may be reattempted at some point.

PR #63, opened 2020-12-01
A couple of GOTK-conformance tweaks

squaremo:
1. Use the probes package from flucd/pkg for installing health checks
 2. Omit suspended objects from metrics (I checked what kustomize-controller does)


relu (approved):
Makes sense.

PR #64, opened 2020-12-01
Make the ImageRepositoryStatus empty when empty

squaremo:
The zero value of an ImageRepositoryStatus includes a struct value in LastScanResult, which prevents it from being omitted when serialised. The effect is that a freshly-constructed ImageRepository will serialise like this:

```yaml
apiVersion: image.toolkit.fluxcd.io/v1alpha1
kind: ImageRepository
metadata:
  name: foobar
  namespace: flux-system
spec:
  image: alpine
  interval: 1m0s
status:
  lastScanResult:
    tagCount: 0
```

This commit changes the LastScanResult field to a pointer type, so it can be `nil` when not set.

The ScanTime field is changed from a pointer to a value, since it should be present in any scan result.


relu (approved):
Looks good! :ok_hand:

Issue #12, opened 2020-07-28
Store image metadata persistently


squaremo:
The image metadata (tags, for now) used by the image reflector are just kept in a hash, to start with. But given the expense (and latency) of scanning for more image metadata, this will soon be a pain point, because any restart will lose everything.

A couple of design ideas:
 - put it in an SQLite DB on a persistent volume (will it fail over nicely? probably not, but decent as a solution for recovery after crashes)
 - put it in a central database which is set up separately (extra for experts ..)

The one thing memcached was quite useful for, in theory, was expiring entries -- any solution here will probably need a garbage collection process.

stefanprodan:
I propose we use [dgraph-io/badger](https://github.com/dgraph-io/badger) that has ACID and TTL support.

squaremo:
> I propose we use [dgraph-io/badger](https://github.com/dgraph-io/badger) that has ACID and TTL support.

😍 where has this database been all my life

squaremo:
@bigkevmcd Neat! That would be a solid step forward.

Yes, the database interfaces are totally minimal at the minute -- there just as placeholders, really. Feel free to expand those, even if it's just to add an error return value for now.

TTLs in the [v1 implementation](https://github.com/fluxcd/flux/tree/master/pkg/registry/cache) are used as a kind of indirect garbage collection. Latterly, memcached is used more like a database, and the TTL is used just to allow memcached a chance to relieve memory pressure (like "once this entry is out of date, feel free to discard it if you need to"). This is not a design I would happily recapitulate, if I'm honest :-)

In _this_ implementation, the database is straight-forwardly a database, with entries added and removed according to the custom resources in play, and there is no need for garbage collection. There _may_ still be a place for TTL, since there is still a notion of entries being out of date, though it needs to be figured out what consequence an out-of-date entry should have. For now I would leave TTL out.

relu:
@bigkevmcd I'd be interested in taking over the badger integration in case you no longer have the time to tackle it. Not sure if you made any progress on it or not, let's catch up and let me know if I can help with anything if you still want to do it.

stefanprodan:
Implemented in #65

PR #65, opened 2020-12-03
Add BadgerDB as a persistent store for the tag data.

bigkevmcd:
This introduces a key/value store for tags, with the keys being the
images and the values being JSON encoded versions of the tags.

The Badger data is stored in a PVC.

Related to #12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN go mod download
 # copy source code
 COPY main.go main.go
 COPY controllers/ controllers/
+COPY internal/ internal/
 
 # build without giving the arch, so that it gets it from the machine
 RUN CGO_ENABLED=0 go build -a -o image-reflector-controller main.go
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -59,6 +59,10 @@ spec:
         volumeMounts:
           - name: temp
             mountPath: /tmp
+          - name: data
+            mountPath: /data
       volumes:
         - name: temp
+          emptyDir: {}
+        - name: data
           emptyDir: {}
--- a/controllers/database.go
+++ b/controllers/database.go
@@ -16,30 +16,16 @@ limitations under the License.
 
 package controllers
 
-import (
-	"sync"
-)
-
-type database struct {
-	mu       sync.RWMutex
-	repoTags map[string][]string
+// DatabaseWriter implementations record the tags for an image repository.
+type DatabaseWriter interface {
+	SetTags(repo string, tags []string) error
 }
 
-func NewDatabase() *database {
-	return &database{
-		repoTags: map[string][]string{},
-	}
-}
-
-func (db *database) Tags(repo string) []string {
-	db.mu.RLock()
-	tags := db.repoTags[repo]
-	db.mu.RUnlock()
-	return tags
-}
-
-func (db *database) SetTags(repo string, tags []string) {
-	db.mu.Lock()
-	db.repoTags[repo] = tags
-	db.mu.Unlock()
+// DatabaseReader implementations get the stored set of tags for an image
+// repository.
+//
+// If no tags are availble for the repo, then implementations should return an
+// empty set of tags.
+type DatabaseReader interface {
+	Tags(repo string) ([]string, error)
 }
--- a/controllers/imagepolicy_controller.go
+++ b/controllers/imagepolicy_controller.go
@@ -49,10 +49,6 @@ import (
 // from.
 const imageRepoKey = ".spec.imageRepository.name"
 
-type DatabaseReader interface {
-	Tags(repo string) []string
-}
-
 // ImagePolicyReconciler reconciles a ImagePolicy object
 type ImagePolicyReconciler struct {
 	client.Client
@@ -204,7 +200,10 @@ func (r *ImagePolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // ---
 
 func (r *ImagePolicyReconciler) calculateLatestImageSemver(pol *imagev1alpha1.ImagePolicyChoice, canonImage string) (string, error) {
-	tags := r.Database.Tags(canonImage)
+	tags, err := r.Database.Tags(canonImage)
+	if err != nil {
+		return "", fmt.Errorf("failed to read images for %q: %w", canonImage, err)
+	}
 	constraint, err := semver.NewConstraint(pol.SemVer.Range)
 	if err != nil {
 		// FIXME this'll get a stack trace in the log, but may not deserve it
@@ -231,7 +230,7 @@ func (r *ImagePolicyReconciler) imagePoliciesForRepository(obj handler.MapObject
 		r.Log.Error(err, "failed to list ImagePolicy for ImageRepository")
 		return nil
 	}
-	reqs := make([]reconcile.Request, len(policies.Items), len(policies.Items))
+	reqs := make([]reconcile.Request, len(policies.Items))
 	for i := range policies.Items {
 		reqs[i].NamespacedName.Name = policies.Items[i].GetName()
 		reqs[i].NamespacedName.Namespace = policies.Items[i].GetNamespace()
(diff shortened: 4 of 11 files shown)


config/manager/deployment.yaml
```diff
@@ -59,6 +59,11 @@ spec:
         volumeMounts:
```
relu, line 59: I think we'll also need to come up with a better default value for the memory request, considering we now run an embedded database as well. Not sure what the actual footprint would be.

relu (changes requested):

relu:
Looks promising. I'm sure it'll have more impact once we start implementing the metadata extraction logic and store that in the database as well.

You'll need to amend your commit and signoff to fix DCO validation.

config/manager/deployment.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
+          fsGroup: 1337
```
relu, line 28: You need to set this at the pod level otherwise it won't work. See info [here](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container):
> Container settings do not affect the Pod's Volumes.

controllers/database.go
```diff
@@ -20,26 +20,27 @@ import (
 	"sync"
```
relu, line 20: Maybe these should be moved to the `suite_test.go` considering these are only used for stubbing purposes.

bigkevmcd: I wonder if they shouldn't go into internal/database ?

relu: Not sure, my first instinct was that they should be close to the tests where they are being used in. But you are right, this is an actual implementation that can even be an alternative to badger. With that in mind, I wonder if we should support toggling between the two implementations i.e. allow the user to configure using either the in memory ephemeral database storage or the persistent badger storage?

bigkevmcd: Adding toggling feels like YAGNI just now, I can see a use-case for it, but it works with the simple image -> tags storage.

I'll move the in-memory version to the suite_test just now, but it's open to seeing it as a togglable replacement

relu (approved):
LGTM :100:

@squaremo you might wanna have a look as well, see if this makes sense to expand for the use cases you had in mind (image metadata).

Dockerfile
```diff
@@ -10,6 +10,7 @@ COPY go.sum go.sum
 # Copy this, which should not change often; and, needs to be in place
 # before `go mod download`.
 COPY api/ api/
+COPY internal/ internal/
```
squaremo, line 13: This can go later, alongside `COPY controllers`; `api/` is copied early just because it has a go.mod and the next stage downloads modules (in the hope that layer can be cached if the go.mods haven't changed).

config/default/pv_claim.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
squaremo, line 1: Allegedly, persistent volumes are slow in some platforms; if this seriously affects the startup time, I would move the persistent volume config to an example (as an optional kustomization, say).

controllers/imagepolicy_controller.go
```diff
@@ -39,17 +39,14 @@ import (
...
 	imagev1alpha1 "github.com/fluxcd/image-reflector-controller/api/v1alpha1"
+	"github.com/fluxcd/image-reflector-controller/internal/database"
 )
 
 // this is used as the key for the index of policy->repository; the
 // string is arbitrary and acts as a reminder where the value comes
 // from.
 const imageRepoKey = ".spec.imageRepository.name"
 
-type DatabaseReader interface {
```
squaremo, line 49: Interfaces generally belong at the location of the requirement, rather than that of the (an) implementation. If it feels off to use a definition in `imagepolicy_controller.go` from `imagerepository_controller.go`, you could put both the definitions in database.go (back in this directory).

internal/database/badger_test.go
```diff
@@ -0,0 +1,110 @@
+/*
+Copyright 2020 The Flux CD authors.
```
squaremo, line 2:
```suggestion
Copyright 2020 The Flux authors.
```

internal/database/badger.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2020 The Flux CD authors.
```
squaremo, line 2:
```suggestion
Copyright 2020 The Flux authors.
```

main.go
```diff
@@ -70,6 +73,7 @@ func main() {
 	flag.BoolVar(&logJSON, "log-json", false, "Set logging to JSON format.")
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	flag.StringVar(&databasePath, "database-path", "/badger", "Where to store the persistent database of image metadata")
```
squaremo, line 76: Better to have this default to something that'll be sensible if you run it in-place, like `/tmp/badger`. For use in a container with a persistent volume, the argument can be patched.

config/default/pv_claim.yaml
stefanprodan: We should use an EmptyDir volume here like we do for source-controller, PVC config could be made available in flux install/bootstrap.

main.go
stefanprodan: Please use the same arg name and location as source-controller https://github.com/fluxcd/source-controller/blob/main/config/manager/deployment.yaml#L43

stefanprodan:
I propose we drop the in-memory store and use Badger with an EmptyDir by default. If people want in-memory storage they can switch the EmptyDir from disk to ram.

internal/database/badger.go
relu: Without the ending dot `.` :smile:

bigkevmcd:
> I propose we drop the in-memory store and use Badger with an EmptyDir by default. If people want in-memory storage they can switch the EmptyDir from disk to ram.

The one area it's arguably useful is in tests...depends on how much isolation you want.

controllers/imagepolicy_controller.go
bigkevmcd: I'm not sure I totally agree with this, while I can understand your thinking, and it comes directly from the Go Code Review guide https://github.com/golang/go/wiki/CodeReviewComments#interfaces and I'd argue that it's useful

It also feels like BadgerDatabase is defining methods on a value that aren't meeting any requirement, and the only place that requirement is defined is in the controller, which I personally find odd.

https://rakyll.org/style-packages/ says...

> In some cases, you may not be able to provide all related types from a single package. It might be noisy to do so, or you might want to publish concrete implementations of a common interface from a separate package, or those types could be owned by a third-party package. 

Which implies that it's common to publish implementations with a common interface in the single package.

squaremo: > It also feels like BadgerDatabase is defining methods on a value that aren't meeting any requirement, and the only place that requirement is defined is in the controller

Yes, it feels like a trapeze act, I know what you mean. It's a shame the CodeReviewComments wiki page doesn't really explain the _why_ of its advice. The way I think about it is this: you use an interface to decouple your use of a value from wherever the value comes from. If you put the interface with the implementation, you reintroduce that coupling.

Thanks for harmonising with my preferences here 🤓

controllers/suite_test.go
```diff
@@ -120,4 +129,6 @@ var _ = AfterSuite(func() {
 	By("tearing down the test environment")
 	err := testEnv.Stop()
 	Expect(err).ToNot(HaveOccurred())
+	Expect(badgerDB.Close()).ToNot(HaveOccurred())
```
squaremo, line 132: Slightly easier to read is

    Expect(badgerDB.Close()).To(Succeed())

squaremo (approved):
This all looks good to me. I'll give it a quick rebass :fish:

PR #75, opened 2021-01-11
Implement policy-level tag regex filtering

### LINKED ISSUE #73: Regular expression filtering
People occasionally want to partition their container image repo and use only a subset of the images. For example, by giving all the images destined for a dev environment a prefix `dev-`.

To give those people a way to filter for the desired subset of images, we can let them supply a regular expression in their `ImagePolicy` spec, and consider only the tags that match the regular expression.

They may also want to use only a portion of the tag for sorting, e.g., if the sort order is CalVer but there is a prefix, like `dev-2021-01-11.build-1`. Therefore it's useful to be able to sort using a particular capture group from the regular expression, or more generally, a replacement expression.

**Why not globbing?**

Globs are simpler to understand -- `dev-*` -- but suffer from a couple of problems:

 1. Libraries don't support captures, so the second requirement above would be tricky;
 2. They are not as general as regular expressions, e.g., it's not possible to filter for things that are CalVer-tagged.

If we implemented globs, we would very likely be asked to also implement regexp -- this is what happened for Flux v1. In the context of image tags, which don't have path separators (so no `**` glob pattern), it is not difficult to construct a regular expression that's equivalent to a glob: `dev-.*` for `dev-*` and so on.

**Place in the API**

https://github.com/fluxcd/image-reflector-controller/pull/58#issuecomment-736544317 and the comment following it https://github.com/fluxcd/image-reflector-controller/pull/58#issuecomment-736652343 suggest that filtering is independent of the ordering. It can either appear in each policy:

```yaml
spec:
  policy:
    semver:
      range: 1.x
      filterRegex: 'dev-v(.*)'
      sortOn: '\1'
```

or it can appear alongside the policy:

```yaml
spec:
  filterRegex: 'dev-v(.*)'
  sortOn: '\1'
  policy:
    semver:
      range: 1.x
```

There's a sense in which the filtering takes place _before_ the policy (e.g., extract the version portion of the tag before checking the semver range), so the latter feels closer to how people might think about it.

relu:
Tag regex filtering allows the user to filter tags based on a regular
expression pattern and enables tag version extraction through capture
group replacement reference.

Fixes #73 

Example use cases (for documentation):

```yaml
filterTags:
  pattern: '^v'
```

```yaml
filterTags:
  pattern: '^nightly-.*'
```

```yaml
filterTags:
  pattern: '^dev-v(.*)'
  extract: '$1'
```

```yaml
filterTags:
  pattern: '^dev-(?P<version>[0-9]+)-alpha.*'
  extract: '$version'
```

```yaml
filterTags:
  pattern: '^rel-(\d+)$'
  extract: '$1'
```

etc.


internal/policy/filter.go
```diff
@@ -0,0 +1,70 @@
+/*
```
relu, line 1: Not super happy about how this was laid out, couldn't think of a nicer way to implement it at the moment though.

api/v1alpha1/imagepolicy_types.go
```diff
@@ -36,6 +36,11 @@ type ImagePolicySpec struct {
 	// selecting the most recent image
 	// +required
 	Policy ImagePolicyChoice `json:"policy"`
+	// FilterTags enables filtering for only a subset of tags based on a set of
+	// rules. If no rules are provided, all the tags from the repository will be
+	// ordered and compared.
+	// +optional
+	FilterTags *TagFilter `json:"filterTags,omitempty"`
```
relu, line 43: Used `filterTags` instead of the proposed `filter` so it's more clear to the user what exactly we are filtering here.

stefanprodan:
@relu can this be used for automating Minio releases? https://hub.docker.com/r/minio/minio/tags?page=1&ordering=last_updated note that users will want the latest multi-arch image (doesn't end in arm/amd/etc) and only the stable releases that start with RELEASE.

internal/policy/filter.go
```diff
@@ -0,0 +1,70 @@
...
+
+	Regexp  *regexp.Regexp
+	Replace string
+}
+
+// NewRegexFilter constructs new RegexFilter object
+func NewRegexFilter(pattern string, replace string) (*RegexFilter, error) {
+	m, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, fmt.Errorf("invalid regular expression pattern '%s': %s", pattern, err.Error())
```
squaremo, line 36:
```suggestion
		return nil, fmt.Errorf("invalid regular expression pattern '%s': %w", pattern, err)
```

internal/policy/filter.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Apply will construct the filtered list of tags based on the provided list of tags
+func (f *RegexFilter) Apply(list []string) {
+	f.filtered = map[string]string{}
+	for _, item := range list {
+		if f.Regexp.MatchString(item) {
+			tag := item
+			if f.Replace != "" {
+				tag = f.Regexp.ReplaceAllString(item, f.Replace)
```
squaremo, line 51: I think you can do this without repeating work, using Find*Index and Expand.

relu:
@stefanprodan yes, since it uses the RFC3339 format it should be lexicographically sortable, using the alphabetical ordering policy it would go something like this:

```yaml
filterTags:
  pattern: '^RELEASE\.(?P<timestamp>.*)Z$'
  extract: '$timestamp'
policy:
  alphabetical:
    order: asc
```

I think it can even be simplified to this:
```yaml
filterTags:
  pattern: '^RELEASE.*'
policy:
  alphabetical:
    order: asc
```

controllers/imagepolicy_controller.go
```diff
@@ -127,7 +127,15 @@ func (r *ImagePolicyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
-			latest, err = policer.Latest(tags)
+			var filter *policy.RegexFilter
+			if pol.Spec.FilterTags != nil {
+				filter, err = policy.NewRegexFilter(pol.Spec.FilterTags.Pattern, pol.Spec.FilterTags.Extract)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				filter.Apply(tags)
+			}
+			latest, err = policer.Latest(tags, filter)
```
squaremo, line 138: Running the flow of control the other way would be clearer here. This way around, you supply a filter to the `Policer`, and it has to know how to deal with it. But if you filter here, and give the `Policer` the extracted names, it can determine which one it wants without having to know whether they are filtered or not. Then you can map back to the original again here (if there was a filter in the first place).

squaremo: One issue, whichever way around you do it, is what to do if the extracted strings are duplicated? E.g., if you have `(dev|stage)-(.*)`, and extract `$2`, you will get a duplicate from the tags `{"dev-v1.0", "stage-v1.0"}`. At present the last one encountered wins (since RegexpFilter keeps track of them in a map). But this isn't stable -- you can get different results, without anything actually changing.

internal/policy/filter.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2020 The Flux authors
```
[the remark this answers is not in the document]
relu: Yes, I didn't bother with this because I already had this laying around since last year. I think 2021 is better for this particular file because it was pretty much rewritten entirely today.

controllers/imagepolicy_controller.go
relu: Thanks for the feedback, I think your suggestion makes a lot of sense and would simplify things significantly in the policies `Latest` implementation.

As for the second comment, I thought about this as well when writing the tests, I wasn't able to think of a real-world scenario where this would become an issue but I think it does make sense to specify this in the documentation since it's indeed an edge case some might stumble upon. I don't know if we can do anything about this conflict aside from warning users about it.

internal/policy/filter.go
relu: Not sure I follow here, what do you refer to as repeated work? I would assume it'll look kind of like this, right?
```go
if submatches := f.Regexp.FindStringSubmatchIndex(item); len(submatches) > 0 {
	tag := item
	if f.Replace != "" {
		result := []byte{}
		result = f.Regexp.ExpandString(result, f.Replace, item, submatches)
		tag = string(result)
	}
	f.filtered[tag] = item
}
```

The above works just fine but I'm not sure I see the benefit :confused:

nomeelnoj:
How would this work if you were adding a build ID to the end? For example, you have two images:

```
dev-v1.2.0-37
dev-v1.2.0-38
```
And your policy is defined as:
```
filterTags:
  pattern: '^dev-v(.*)-[0-9]+'
  extract: '$1'
policy:
  semver:
    range: 1.2.x
```

Would the `dev-v1.2.0-38` image get deployed because it is newer than the previous? or would the reflector controller ignore this image because it already has one that matches the filter?

relu:
@nomeelnoj you will want to include the pre-release part (what you mentioned as being the build id) to the capture group.

```yaml
filterTags:
  pattern: '^dev-v(.*)'
  extract: '$1'
policy:
  semver:
    range: 1.2.x-0
```

Notice that I've added a `-0` to your range, this is to instruct semver evaluation to consider the pre-release part of the version as well.

nomeelnoj:
@relu thanks! very excited for this to be merged, as I believe this is the final blocker from getting us over to v2 (v1 is less than functional because the helm operator is now having issues with the old stable repo being deprecated).  How long after this is merged will a release be cut?

stefanprodan (approved):
LGTM

Thanks @relu 🥇

internal/policy/filter.go
squaremo: Both `MatchString` and `ReplaceAllString` "run" the regular expression -- it's just that the former disregards submatches, and the latter _also_ does substitution. On the other hand, `FindStringSubmatchIndex` runs the regular expression to get the submatches (and whether it matches at all), but ExpandString does only the substitution work.

It is unlikely to make a vast difference to any individual match, but all else being equal, it's worth avoiding extra calculation.

squaremo (approved):
A very welcome addition to the API, thank you @relu.

controllers/imagepolicy_controller.go
squaremo: > it does make sense to specify this in the documentation since it's indeed an edge case some might stumble upon

I agree with this, and

> warning users about it

with this. If it happens, it could be pretty mysterious, so we could log a message with a reference to docs or decent search term.

squaremo: (we can discuss these alternatives a bit more in https://github.com/fluxcd/image-reflector-controller/issues/77)

internal/policy/filter.go
relu: Got it, thanks for the explanation, I just looked into more detail in the implementation of `ReplaceAllString` vs `ExpandString` and I did notice now what you mean. I will refactor.

squaremo:
I'll take a look at those test failures ...

PR #78, opened 2021-01-13
Avoid race between DB reads and DB close in tests

squaremo:
Badger will panic with a nil pointer dereference if you do a transaction after a database is closed. In the controller test suite, the database is closed but the controllers, which read from it, are left running.

To remove this race, it is necessary to tell the controller manager explicitly to shut down; this is done by cancelling the context given to it. This isn't synchronous, because the manager is running in its own goroutine, and cancelling a context can't block until all the consequences are done -- but it may be good enough for our purposes. (If not, we can add some shutdown protocol to the controllers, e.g., channels which they close when exiting).

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


hiddeco (approved):
:100:

relu (approved):

Issue #82, opened 2021-01-16
Controller does not start on RaspberryPI (unable to open the Badger database)


bjornmagnusson:
Running flux (0.6.1) in a cluster using two RasberryPI. 
All but image-reflector-controller starts up fine.
Image-reflector-controller fails to allocate memory for the Badger database.

Logs:
```
kubectl -n flux-system logs image-reflector-controller-6bbbcc5c76-ttwcq
{"level":"info","ts":"2021-01-16T10:10:43.238Z","logger":"controller-runtime.metrics","msg":"metrics server is starting to listen","addr":":8080"}
badger 2021/01/16 10:10:43 INFO: All 0 tables opened in 0s
{"level":"error","ts":"2021-01-16T10:10:43.302Z","logger":"setup","msg":"unable to open the Badger database","error":"Mmap value log file. Path=/data/000000.vlog. Error=cannot allocate memory"}
```

Flux-system pods
```
kubectl get pods -n flux-system -owide
NAME                                          READY   STATUS             RESTARTS   AGE   IP           NODE   NOMINATED NODE   READINESS GATES
source-controller-bb458c944-kj76r             1/1     Running            0          15h   10.42.0.24   rpi1   <none>           <none>
helm-controller-77c9c759fb-z5f95              1/1     Running            10         15h   10.42.1.41   rpi2   <none>           <none>
notification-controller-6fbbc586b7-4kkff      1/1     Running            10         15h   10.42.1.42   rpi2   <none>           <none>
kustomize-controller-64b894fb67-8fnkq         1/1     Running            10         15h   10.42.1.44   rpi2   <none>           <none>
image-automation-controller-cd59d8b74-hh82w   1/1     Running            10         15h   10.42.1.43   rpi2   <none>           <none>
image-reflector-controller-5d98f74bc-6sxdr    0/1     CrashLoopBackOff   7647       27d   10.42.1.40   rpi2   <none>           <none>
image-reflector-controller-6bbbcc5c76-ttwcq   0/1     CrashLoopBackOff   186        15h   10.42.1.45   rpi2   <none>           <none>
```
Flux:
`flux --version
flux version 0.6.1
`

Cluster memory/cpus
```
kubectl top node
NAME   CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
rpi1   712m         17%    793Mi           81%       
rpi2   491m         12%    619Mi           63%
```

Raspberry PI details:
```
uname -a
[...]

PR #84, opened 2021-01-16
Update Badger to v3.2011.0

stefanprodan:
We were using Badger v1 while the latest version is v3. This PR upgrades Badger to v3.2011.0 that should fix the `cannot allocate memory` error on RaspberryPI (ref: #82).


relu (approved):
Looks to be a pretty smooth upgrade :+1:

hiddeco (approved):

PR #85, opened 2021-01-16
Release v0.2.1

stefanprodan:


relu (approved):

hiddeco (changes requested):
> This release is not backward compatible with Badger v2.x.x

from https://github.com/dgraph-io/badger/releases/tag/v3.2011.0

If this means that the storage isn't backwards compatible, the new version should be `v0.3.0` given persistent storage options.

PR #90, opened 2021-01-20
Add examples to the API spec docs

stefanprodan:


docs/spec/v1alpha1/imagepolicies.md
```diff
@@ -124,5 +124,73 @@ The `LatestImage` field contains the image selected by the policy rule, when it
 There is one condition that may be present: the GitOps toolkit-standard `ReadyCondition`. This will
 be marked as true when the policy rule has selected an image.
 
-[image-automation-controller]: https://github.com/image-automation-controller
```
squaremo, line 127: Are these not supported by the markdown engine?

docs/spec/v1alpha1/imagepolicies.md
```diff
@@ -124,5 +124,73 @@ The `LatestImage` field contains the image selected by the policy rule, when it
...
-[semver-range]: https://github.com/Masterminds/semver#checking-version-constraints
+## Examples
```
relu, line 127: It might make sense to send people to the [regex syntax doc page](https://golang.org/pkg/regexp/syntax/) for help with compiling supported regular expression patterns.
In the future, once we can develop validation webhooks, we'll be able to provide feedback to the user if the regular expression is incorrect.

docs/spec/v1alpha1/imagepolicies.md
```diff
@@ -124,5 +124,73 @@ The `LatestImage` field contains the image selected by the policy rule, when it
...
+
+Select the latest `dev` branch build tagged as `<BRANCH>-<BUILD-ID>` (alphabetical):
+
+```yaml
+kind: ImagePolicy
+spec:
+  filterTags:
+    pattern: '^dev-(?P<id>.*)'
+    extract: '$id'
```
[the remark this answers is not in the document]
squaremo: Is this meaningfully different to the Minio example?

docs/spec/v1alpha1/imagepolicies.md
squaremo: The same mechanism is used in the other file, so I would guess yes it is supported ..

squaremo (approved):
<3 examples, thank you Stefan

docs/spec/v1alpha1/imagepolicies.md
stefanprodan: I can add them back, we don't use links like this in the other docs, no idea if they work with mkdocs or not.

docs/spec/v1alpha1/imagepolicies.md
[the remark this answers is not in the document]
stefanprodan: I've seen this pattern in many setups `<branch>-<build-id>` that's why I put it here. I can remove it if you want, it is similar to Minio but easier to grok.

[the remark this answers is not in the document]
squaremo: OK yeah -- sounds like it's meaningfully different because it's a use seen in the wild, and not obviously the same as using CalVer, even though it technically works the same.

docs/spec/v1alpha1/imagepolicies.md
squaremo: It's standard markdown, so I would expect it does work. I use them for keeping paragraph text clear of distracting full-line URLs.

docs/spec/v1alpha1/imagepolicies.md
stefanprodan: I've added the link to Go docs. Thanks @relu for the suggestion.

docs/spec/v1alpha1/imagepolicies.md
stefanprodan: I added them back.

relu (approved):

PR #95, opened 2021-01-22
Update BadgerDB to v3.2011.1

stefanprodan:
Includes fixes for ARM 32bit build that we've been using from master


relu (approved):
LGTM :heavy_check_mark:

PR #104, opened 2021-02-11
Implement numerical ordering policy

### LINKED ISSUE #102: Extend image policy with numerical ordering
While alphabetical ordering works great with image tags that contain a RFC3339 timestamp, it's not suitable for CI build IDs, as these are numbers with no padding. 

I propose we introduce a new ordering option called `numerical` to be able to correct detect the latest build for tags in the format `<PREFIX>-<BUILD_ID>`. 

Example:

```yaml
kind: ImagePolicy
spec:
  filterTags:
    pattern: '^main-[a-fA-F0-9]+-(?P<id>.*)'
    extract: '$id'
  policy:
    numerical:
      order: asc
```

Given the tags `main-845d3a80-100` and `main-3e32dc8a-2` the `numerical` policy, unlike  `alphabetical`, will chose `main-3e32dc8a-100` as the latest build.

relu:
Converts the given list of tags to floats and compares them based on the
ordering rule.
If a tag is not convertible, it will err and fail to compute the latest
version.

Fixes #102


internal/policy/numericall_test.go
```diff
@@ -0,0 +1,147 @@
+/*
```
hiddeco, line 1: The name of this file does not match the name of the file it tests (`numerical` vs `numericall`).

relu: And here I was thinking it was all perfect and fine. Thanks!

stefanprodan (approved):
LGTM

Thanks @relu 🥇 

I can do a followup PR to update the API spec docs and examples.

config/crd/bases/image.toolkit.fluxcd.io_imagepolicies.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.3.0
```
stefanprodan, line 7: @relu can you delete the `controller-gen` binary from your local Go bin and rerun `make test` this will pull v0.4.1 and undo the above.

relu: Should be good now. Thanks!

PR #106, opened 2021-02-11
Add numerical policy to API docs

stefanprodan:


relu (approved):
LGTM :heavy_check_mark:

PR #112, opened 2021-02-24
Release v0.6.1

stefanprodan:


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## 0.6.1
```
hiddeco, line 3: Wondering if, given that the way we "package" our release manifests has changed, all releases should be a new MINOR for semver predictability/correctness.

PR #119, opened 2021-03-12
Fix "last scan" print column for ImageRepository

### LINKED ISSUE #118: ImageRepository not populating Last Scan in additionalPrinterColumns
The additionalPrinterColumn for LAST SCAN is not populating.  I believe it was #56 that introduced this.  I have confirmed that updating the CRD as follows populates the field:

https://github.com/fluxcd/image-reflector-controller/blob/main/config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml#L20

should instead read 
```
.status.lastScanResult.scanTime
```

relu:
Fixes #118


stefanprodan (approved):
LGTM

Thanks @relu

PR #123, opened 2021-03-17
Record suspension metrics

SomtochiAma:
Part of (flux#590)[https://github.com/fluxcd/flux2/issues/590]

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -88,6 +88,8 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	defer r.recordSuspension(ctx, imageRepo)
+
 	log := logr.FromContext(ctx)
 
 	if imageRepo.Spec.Suspend {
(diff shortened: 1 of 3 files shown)


controllers/imagerepository_controller.go
```diff
@@ -411,3 +413,22 @@ func (r *ImageRepositoryReconciler) recordReadinessMetric(ctx context.Context, r
 		}, !repo.DeletionTimestamp.IsZero())
 	}
 }
+
+func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageRepo imagev1alpha1.ImageRepository) {
```
squaremo, line 417: This looks like it depends very little on `imagev1alpha1.ImageRepository` -- can it be factored out into fluxcd/pkg, with a very minimal interface as its argument?

stefanprodan: The runtime/metrics package deals with object references, what factored out do you have in mind?

squaremo: This method is near identical to https://github.com/fluxcd/helm-controller/pull/236/files#diff-4a892aba8c8073dd54b0ecdb7493670d3b0486a078d02a83efa0cecc09e09b67R737, https://github.com/fluxcd/notification-controller/pull/164/files#diff-c2f1e669308ef8c3c3b1083764bb88a89f3de34b885d30edeae65d1d53a6136aR115, https://github.com/fluxcd/kustomize-controller/pull/299/files#diff-5d1fa9b5557ebd96e4511008d93cb5cc21084d81f6baac4d1ebbbfd28a69eba0R834, etc. The only way in which it differs is the type of the receiver (from which only the metrics recorder and scheme are actually used, and those are common to all the controllers), and the API type, which is only required to have a .Spec.Suspend field, implement whatever interface `reference.GetReference()` accepts, and have a deletion timestamp (which is already available via `GetDeletionTimestamp()` in `metav1.ObjectMeta`, which is embedded by all the API types).

stefanprodan: I wander if we could factored out the readiness recorder as well, the Ready condition is the same for all our custom resources.

squaremo: I reckon so -- I had an attempt in https://github.com/fluxcd/pkg/pull/56 but didn't line things up quite right, and was trying to do too much at once. A decent start would be simply to provide funcs `RecordSuspension` and `RecordReadiness`; a stretch might be to make these methods of a struct (possibly MetricsRecorder).

squaremo: wdyt @SomtochiAma, are you up for this refactor?

Issue #124, opened 2021-03-22
unable to find docker auth in secret


HarryGogonis:
The ImageRepository uses `ref.Context().RegistryStr()` to get the container registry auth from the map in the secret
https://github.com/fluxcd/image-reflector-controller/blob/e164cfc676cebc0b509cc888dfd3b84964e03d54/controllers/imagerepository_controller.go#L193
https://github.com/fluxcd/image-reflector-controller/blob/e164cfc676cebc0b509cc888dfd3b84964e03d54/controllers/imagerepository_controller.go#L368

For docker, this value is `index.docker.io`. Unfortunately, many .dockerconfigjson contain the full domain `https://index.docker.io/v2` as [shown in the k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)

I can change my docker secret for now, but it might be a "gotcha" for other users

squaremo:
Hmm yeah, we had to deal with this in Flux v1 too. The way [it is done there](https://github.com/fluxcd/flux/blob/master/pkg/registry/credentials.go#L71) is to parse each map key from `auths`  as though it's a URL, and use only the hostname. There's a bit of fiddliness since the map key might _not_ be a URL, but it works well enough.

PR #130, opened 2021-04-08
Parse docker auths and use only hostname

SomtochiAma:
Fix: #124 
Fix: #131

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -22,8 +22,11 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -70,6 +73,10 @@ type ImageRepositoryReconciler struct {
 		DatabaseWriter
 		DatabaseReader
 	}
+}
+
+type dockerConfig struct {
+	Auths map[string]authn.AuthConfig
 }
 
 // +kubebuilder:rbac:groups=image.toolkit.fluxcd.io,resources=imagerepositories,verbs=get;list;watch;create;update;patch;delete
@@ -190,7 +197,7 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1
 			)
 			return err
 		}
-		auth, err := authFromSecret(authSecret, ref.Context().RegistryStr())
+		auth, err := authFromSecret(authSecret, ref)
 		if err != nil {
 			imagev1alpha1.SetImageRepositoryReadiness(
 				imageRepo,
@@ -355,17 +362,21 @@ func (r *ImageRepositoryReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // authFromSecret creates an Authenticator that can be given to the
 // `remote` funcs, from a Kubernetes secret. If the secret doesn't
 // have the right format or data, it returns an error.
-func authFromSecret(secret corev1.Secret, registry string) (authn.Authenticator, error) {
+func authFromSecret(secret corev1.Secret, ref name.Reference) (authn.Authenticator, error) {
 	switch secret.Type {
 	case "kubernetes.io/dockerconfigjson":
-		var dockerconfig struct {
-			Auths map[string]authn.AuthConfig
-		}
+		var dockerconfig dockerConfig
 		configData := secret.Data[".dockerconfigjson"]
 		if err := json.NewDecoder(bytes.NewBuffer(configData)).Decode(&dockerconfig); err != nil {
 			return nil, err
 		}
-		auth, ok := dockerconfig.Auths[registry]
+
+		authMap, err := parseAuthMap(dockerconfig)
+		if err != nil {
+			return nil, err
+		}
+		registry := ref.Context().RegistryStr()
+		auth, ok := authMap[registry]
 		if !ok {
 			return nil, fmt.Errorf("auth for %q not found in secret %v", registry, types.NamespacedName{Name: secret.GetName(), Namespace: secret.GetNamespace()})
 		}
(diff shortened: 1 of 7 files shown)


controllers/imagerepository_controller.go
```diff
@@ -432,3 +444,46 @@ func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageR
...
+
+		host = u.Host
+
+		auth[host] = entry
+	}
+
+	return auth, nil
+}
+
+func parseURL(urlStr string) (*url.URL, error) {
```
relu, line 464: I would reverse the logic here so that `url.Parse()` is only called once. Something like this:
```
if !strings.HasPrefix(urlStr, "http://") || !strings.HasPrefix(urlStr, "https://") {
  urlStr = fmt.Sprintf("https://%s/", urlStr)
}
u, err := url.Parse(urlStr)
```
Checking the format of the string first (if it contains the scheme) and parsing only after prepending the scheme (if it is necessary).

controllers/imagerepository_controller.go
```diff
@@ -432,3 +444,46 @@ func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageR
...
+	// if anything went wrong try to prepend https://
+	if err != nil || u.Host == "" {
+		u, err = url.Parse(fmt.Sprintf("https://%s/", urlStr))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if u.Host == "" {
+		return nil, errors.New(fmt.Sprintf("Invalid registry auth url: %s. Must be a valid http address (e.g. https://gcr.io/v1/)", urlStr))
```
relu, line 485:
```suggestion
		return nil, errors.New(fmt.Sprintf("Invalid registry auth url: %s. Must be a valid URI (e.g. https://gcr.io/v1/)", urlStr))
```

controllers/secret_test.go
```diff
@@ -49,3 +49,28 @@ func TestExtractAuthn(t *testing.T) {
...
+	b, err := ioutil.ReadFile("testdata/secret2.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var secret corev1.Secret
+	if err = json.Unmarshal(b, &secret); err != nil {
+		t.Fatal(err)
+	}
+
+	tests := []string{
```
relu, line 63: Maybe add one with `http://` as well?

controllers/imagerepository_controller.go
```diff
@@ -432,3 +444,46 @@ func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageR
...
+	}
+
+	return u, nil
```
relu, line 488: Considering that in all use cases only the `u.Host` attribute ends up being used, it might make sense to just return that and maybe rename the func to `getURLHost`.

relu (changes requested):
Nice work, @SomtochiAma!

I left some comments, hope my suggestions make sense.

squaremo (changes requested):
Thanks for taking this on :+1:
Requested changes:
 - please correct the ignored error, and consider the suggestion I put alongside that;
 - I think it's worth looking at the code comment and error message about URLs, and seeing if they are still accurate

controllers/imagerepository_controller.go
```diff
@@ -358,16 +365,22 @@ func (r *ImageRepositoryReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		}
+		var dockerconfig dockerConfig
 		configData := secret.Data[".dockerconfigjson"]
 		if err := json.NewDecoder(bytes.NewBuffer(configData)).Decode(&dockerconfig); err != nil {
 			return nil, err
 		}
-		auth, ok := dockerconfig.Auths[registry]
+
+		// parse registry so we can look up with the host
+		registryHost, err := getURLHost(registry)
```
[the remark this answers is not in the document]
SomtochiAma: If the registry entry isn't sanitized, The old tests fail ("https://index.docker.io").

controllers/imagerepository_controller.go
```diff
@@ -432,3 +445,44 @@ func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageR
...
+	// ensure url has https:// or http:// prefix
+	if !strings.HasPrefix(urlStr, "http://") && !strings.HasPrefix(urlStr, "https://") {
+		urlStr = fmt.Sprintf("https://%s/", urlStr)
+	}
+
+	// Some users were passing in credentials in the form of
+	// http://docker.io and http://docker.io/v1/, etc.
+	// So strip everything down to the host.
+	// Also, the registry might be local and on a different port.
+	// So we need to check for that because url.Parse won't parse the ip:port format very well.
```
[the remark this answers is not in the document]
SomtochiAma: `url.Parse` still doesn't parse ip:port well if it doesn't start with `http(s)://`.

controllers/imagerepository_controller.go
```diff
@@ -474,14 +478,15 @@ func getURLHost(urlStr string) (string, error) {
...
-	// So we need to check for that because url.Parse won't parse the ip:port format very well.
 	u, err := url.Parse(urlStr)
 	if err != nil {
 		return "", err
 	}
 
 	if u.Host == "" {
-		return "", errors.New(fmt.Sprintf("Invalid registry auth url: %s. Must be a valid URI (e.g. https://gcr.io/v1/)", urlStr))
+		return "", errors.New(fmt.Sprintf(
+			"Invalid registry auth url: %s. Must be a valid hostame(e.g. gcr.io) or url (e.g. https://gcr.io/v1/)",
```
squaremo, line 488: Sorry to be a nitpicker -- it could also be a non-hostname, non-URL, like `index.docker.io/v2` and get through the func successfully. The requirement is that either it's a valid HTTP[S] URL, or adding https:// makes it a valid HTTPS URL.

squaremo: I realise lots of the things I'm commenting on actually come from the code you're borrowing, for which I can only apologise :-)

SomtochiAma: Sure. I will correct that.

SomtochiAma: Is this okay ?
```
Must be a valid http URL (e.g. https://gcr.io/v1/) or prepending with 'https://' makes it a valid http URL
```

controllers/imagerepository_controller.go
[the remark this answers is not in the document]
SomtochiAma: All right. I will correct it and update the tests.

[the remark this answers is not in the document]
SomtochiAma: Will it be better `name.Registry` to the `authFromSecret function` then grab the string  with `.RegistryStr()`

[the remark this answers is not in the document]
squaremo: > Will it be better name.Registry to the authFromSecret function then grab the string with .RegistryStr()

I reckon so, that way the caller is encouraged (by the type system, and by the self-documenting func signature) to provide the right kind of value.

controllers/imagerepository_controller.go
```diff
@@ -432,3 +442,46 @@ func (r *ImageRepositoryReconciler) recordSuspension(ctx context.Context, imageR
...
+	// So strip everything down to the host.
+	// Also, the registry might be local and on a different port.
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return "", err
+	}
+
+	if u.Host == "" {
+		return "", errors.New(fmt.Sprintf(
+			"Invalid registry auth url: %s. Must be a valid http url (e.g. https://gcr.io/v1/) or prepending with 'https' makes it a valid http url",
```
squaremo, line 482: This one is tricky to word isn't it? I'm not happy with my earlier suggestion, now :-S 

How about this:

```suggestion
			"Invalid registry auth key: %s. Expected an HTTPS URL (e.g. 'https://gcr.io/v1/' or 'https://gcr.io'), or the same without the 'https://' (e.g., 'gcr.io/v1/' or 'gcr.io')"
```

squaremo: Blah, I'm just bikeshedding now -- your one is :+1:

stefanprodan: Let's replace GCR with docker.io please, this would look very strange to someone using AWS or Azure.

squaremo: (incorporated that into the click-to-commit)

squaremo (approved):

squaremo:
Thanks for the follow-up, @SomtochiAma 💯

stefanprodan (approved):
LGTM

Thanks @SomtochiAma please squash all commits.

relu (approved):

PR #132, opened 2021-04-19
Move API v1alpha1 to v1alpha2

squaremo:
The image automation part of the API has changed structure (see [1]),
and had a version bump from `v1alpha1` to `v1alpha2`. Since the types
here are also in `image.toolkit.fluxcd.io`, there will be less
complication if they also get a version bump even though they aren't
changing.

For the automation types, leaving the old version `v1alpha1`
definitions in place leads to Kubernetes trying to upgrade objects and
dropping most fields (which have changed location), so they have been
removed. That would not happen with the types here, but for
consistency I am removing the old API version here too.

[1] https://github.com/fluxcd/image-automation-controller/pull/139

NB This will not work for a live system at present; applying the CRD gives the error:

```
The CustomResourceDefinition "imagerepositories.image.toolkit.fluxcd.io" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions
```


stefanprodan (approved):
LGTM

Thanks @squaremo ☕

stefanprodan (approved):
Thanks @hiddeco

PR #144, opened 2021-05-14
Better error reporting for image policy evaluation

relu:
A more verbose message associated with the error should give the user
better understanding of the cause and the nature of the failure.

### DIFF
--- a/internal/policy/numerical.go
+++ b/internal/policy/numerical.go
@@ -66,7 +66,7 @@ func (p *Numerical) Latest(versions []string) (string, error) {
 
 		switch {
 		case i == 0:
-			break // First iteration, nothing to compare
+			// First iteration, nothing to compare
 		case p.Order == NumericalOrderAsc && cv < pv, p.Order == NumericalOrderDesc && cv > pv:
 			continue
 		}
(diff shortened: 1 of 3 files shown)


internal/policy/numerical.go
```diff
@@ -66,7 +66,7 @@ func (p *Numerical) Latest(versions []string) (string, error) {
 
 		switch {
 		case i == 0:
-			break // First iteration, nothing to compare
+			// First iteration, nothing to compare
```
squaremo, line 69: I like the explicit `break`, since it disambiguates for people who have trouble remembering whether Go's `case` has fall-through or not (like me).

squaremo (approved):
A sensible improvement, thanks @relu 🍇

internal/policy/numerical.go
relu: I just went with the linter recommendation. I can definitely revert back to using break if you think it's worth it, no problem.

squaremo: I often find myself disagreeing with linters. Maybe I should consider .... rewriting all the linters ...

squaremo: Not a blocker, I leave it to your judgement @relu :-)

PR #151, opened 2021-06-10
Add nightly builds workflow and allow RC releases

hiddeco:


stefanprodan:
Let's disable the multi-arch build for each PR.

stefanprodan (approved):

squaremo:
It'd be good to see a rationale for this, especially for disabling multi-arch builds.

stefanprodan:
We've moved the multi-arch build to nightly, as the build takes anywhere from 20 to 35 minutes making the release process too slow. Note that this is the only repo where I've enabled multi-arch builds for PRs. This PR brings IRC CI on pair with all the others controllers.

hiddeco:
In the other Flux projects we were only building the multi-arch image during the release, this has (unsurprisingly) resulted in two releases that failed at the last moment.

To mitigate against this, we introduced a "nightly" build that runs once every night and on-demand, so that small PRs and/or things that have little relevance to 32-bit do not burn useless minutes in CI, as the build time is increased by ^5-10 for multi-arch builds.

When I created this PR I was unaware the multi-arch build was already run for the "normal build", I was just trying to get things in-line with our other projects.

hiddeco:
The other option I suggested instead of "nightly" is to run the multi-arch build on merges to `main`, but not for pull requests themselves. That may be more up to your likening.

stefanprodan:
> When I created this PR I was unaware the multi-arch build was already run for the "normal build", I was just trying to get things in-line with our other projects.

Yea it's fault, I should've undo it here after the Badger bug fixes.

squaremo:
OK, thanks for explaining @hiddeco and @stefanprodan. I do prefer the "full build on merge to main" alternative, but nightlies is fine too.

hiddeco:
What if we refactor things in a next iteration so that merges on `main` always build a multi-arch image, while the "nightly" also pushes the image (as we did for Flux in the past so that people have access to "early releases")?

hiddeco:
Merging this to align things, https://github.com/fluxcd/flux2/issues/1498 discusses a future revision.

PR #153, opened 2021-06-23
Make concurrent reconciliation configurable

### LINKED ISSUE #148: Repo timeout locks full reconciliation cycle
Reflector-version: 0.6.0 

```    
Image:         ghcr.io/fluxcd/image-reflector-controller:v0.6.0
    Image ID:      docker-pullable://ghcr.io/fluxcd/image-reflector-controller@sha256:3f2f7e14681e92165b6bcd86fb28cf87bda60628535ab7613a3d3095426cf7b4
    Ports:         8080/TCP, 9440/TCP
    Host Ports:    0/TCP, 0/TCP
    Args:
      --events-addr=http://notification-controller/
      --watch-all-namespaces=true
      --log-level=info
      --log-encoding=json
      --enable-leader-election
      ```
      
  ```
  Please take a look on ts. 15:41:18. -->(~2 minutes) -->  15:43:52 --> 15:43:52
  ```
  {"level":"info","ts":"2021-06-02T15:41:18.217Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"reconciliation finished in 114.811428ms, next run in 1m0s","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":*********","namespace":"flux-system"}
{"level":"error","ts":"2021-06-02T15:43:52.355Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"*********,"namespace":"flux-system","error":"Get \"https://git.something.com:5050/v2/\": dial tcp **********:5050: i/o timeout"}
{"level":"info","ts":"2021-06-02T15:43:52.535Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"reconciliation finished in 179.660488ms, next run in 1m0s","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"********","namespace":"flux-system"}
```

1. Imagerepo reconciliation is stoped in case of timeout for one of the repo. 
In this case, one broken image repo(even temporary) breaks all ci/cd.

2. Another problem (small one, the first point is more relevant) -  by default timeout is equal to the interval. And in theory, an interval for 15 minutes(which is a quite possible scenario) can lock all reconciliation for 15 minutes.
Can it be 1 minute by default to prevent such cases?

makkes:
Default for both, the ImageRepository and the ImagePolicy controllers
is 4 workers.

fix: #148

Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/main.go
+++ b/main.go
@@ -138,14 +138,15 @@ func main() {
 	pprof.SetupHandlers(mgr, setupLog)
 
 	if err = (&controllers.ImageRepositoryReconciler{
-		Client:                  mgr.GetClient(),
-		Scheme:                  mgr.GetScheme(),
-		EventRecorder:           mgr.GetEventRecorderFor(controllerName),
-		ExternalEventRecorder:   eventRecorder,
-		MetricsRecorder:         metricsRecorder,
-		Database:                db,
+		Client:                mgr.GetClient(),
+		Scheme:                mgr.GetScheme(),
+		EventRecorder:         mgr.GetEventRecorderFor(controllerName),
+		ExternalEventRecorder: eventRecorder,
+		MetricsRecorder:       metricsRecorder,
+		Database:              db,
+	}).SetupWithManager(mgr, controllers.ImageRepositoryReconcilerOptions{
 		MaxConcurrentReconciles: concurrent,
-	}).SetupWithManager(mgr); err != nil {
+	}); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", imagev1.ImageRepositoryKind)
 		os.Exit(1)
 	}
@@ -156,8 +157,9 @@ func main() {
 		ExternalEventRecorder: eventRecorder,
 		MetricsRecorder:       metricsRecorder,
 		Database:              db,
+	}).SetupWithManager(mgr, controllers.ImagePolicyReconcilerOptions{
 		MaxConcurrentReconciles: concurrent,
-	}).SetupWithManager(mgr); err != nil {
+	}); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", imagev1.ImagePolicyKind)
 		os.Exit(1)
 	}
--- a/main.go
+++ b/main.go
@@ -68,6 +68,7 @@ func main() {
 		watchAllNamespaces      bool
 		storagePath             string
 		storageValueLogFileSize int64
+		concurrent              int
 	)
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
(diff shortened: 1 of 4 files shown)


relu:
Looking good, @makkes!

One suggestion I have, for consistency's sake, I think it would make sense to define some types named `ImagePolicyReconcilerOptions` and respectively `ImageRepositoryReconcilerOptions` to hold the reconciliation specific options separately, similar to how it's done in [other](https://github.com/fluxcd/source-controller/blob/44dd9d7e28820c2103f541737c3bf60cdc5ba926/controllers/bucket_controller.go#L70-L72) [controllers](https://github.com/fluxcd/helm-controller/blob/893d804e4f130b20703048196184ef9fabfeebc4/controllers/helmrelease_controller.go#L270-L274).
Also see [an example](https://github.com/fluxcd/helm-controller/blob/893d804e4f130b20703048196184ef9fabfeebc4/controllers/helmrelease_controller.go#L84) of the `SetupWithManager` implementation.

makkes:
> Looking good, @makkes!
> 
> One suggestion I have, for consistency's sake, I think it would make sense to define some types named `ImagePolicyReconcilerOptions` and respectively `ImageRepositoryReconcilerOptions` to hold the reconciliation specific options separately

Yeah, I saw these in the other controllers and was wondering why these were different. If you don't mind, I'll introduce these types in a subsequent PR to keep this focussed on the reconciliation configurability. What do you think?

stefanprodan:
@makkes it's ok to introduce the `Options` types in this PR.

makkes:
> @makkes it's ok to introduce the `Options` types in this PR.

Alright, I'll add a commit shortly.

makkes:
@relu @stefanprodan added option types for both controllers.

stefanprodan (approved):
LGTM

Thanks @makkes 🥇 

PS. Can you please do the same in image-automation-controller?

makkes:
>PS. Can you please do the same in image-automation-controller?
sure thing

main.go
```diff
@@ -77,6 +78,7 @@ func main() {
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.StringVar(&storagePath, "storage-path", "/data", "Where to store the persistent database of image metadata")
 	flag.Int64Var(&storageValueLogFileSize, "storage-value-log-file-size", 1<<28, "Set the database's memory mapped value log file size in bytes. Effective memory usage is about two times this size.")
+	flag.IntVar(&concurrent, "concurrent", 4, "The number of concurrent resource reconciles.")
```
squaremo, line 81: I think this flag definition comes from elsewhere; but either way, it deserves a better description, on two counts:

 - it's the _maximum_ number of workers
 - "reconciles" doesn't tell the user what it affects.

Something like "The maximum number of scans that can run concurrently" would be an improvement, I reckon. WDYT?

squaremo (approved):
Nice and neat, and consistent with e.g., kustomize controller, which is appreciated :-)
I've suggested an improvement, which you can take or leave ...

main.go
squaremo: Oh it was merged already. OK never mind.

makkes:
@stefanprodan https://github.com/fluxcd/image-automation-controller/pull/192

main.go
makkes: @squaremo I agree. I'll come up with a follow-up PR.

stefanprodan: We could extract this into `fluxcd/pkg/runtime` like we did with the log level and other common flags. The description used here is the same for all the others controllers, better to change it on place.

PR #154, opened 2021-06-24
Graduate v1alpha2 API to v1beta1

squaremo:
The schema is the same, which means this is a case of

 - copying over the type definitions;
 - changing the "stored" CRD version;
 - giving a CRD conversion of None, meaning just change the apiVersion;
 - using the new API package in the controller code;
 - moving the API docs to v1beta1


config/crd/patches/conversion.yaml
```diff
@@ -0,0 +1,17 @@
...
+  conversion:
+    strategy: None
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: imagepolicies.image.toolkit.fluxcd.io
+spec:
+  conversion:
+    strategy: None
```
stefanprodan, line 17: `None` is the default conversion strategy, no need for this patch, worked fine without it when we promoted v1alpha1 to v1alpha2.

squaremo: OK, if there's no need to be explicit about it, I can remove it -- fewer files innit :-)

stefanprodan (approved):
LGTM

Thanks @squaremo 🏅

PR #162, opened 2021-08-06
[RFC] Add ACL support for allowing cross-namespace access to image repository

stefanprodan:
## Problem

In the current implementation, users are forced into placing all the image automation objects in the same namespace because the `ImagePolicy` can refer to an `ImageRepository` only in the same namespace as itself. This means users have to copy the image pull secrets from their original namespaces into flux-system or copy the Git auth secret in all apps namespaces.

Initial proposal: https://github.com/fluxcd/flux2/pull/582#issuecomment-838725285
Issues: https://github.com/fluxcd/image-automation-controller/issues/85 https://github.com/fluxcd/image-automation-controller/pull/151

## Proposed solution

To allow an `ImagePolicy` to refer an `ImageRepository` across the cluster, the `ImageRepositoryRef` type changes from `LocaObjectReference` to `NamespacedObjectReference`.

```go
type ImagePolicySpec struct {
	// ImageRepositoryRef points at the object specifying the image being scanned
	// +required
	ImageRepositoryRef meta.NamespacedObjectReference `json:"imageRepositoryRef"`
}
```

The `ImageRepositoryRef.Namespace` is optional, when not specified, the controller behaves like before: it looks for the 
`ImageRepository` in the same namespace as the `ImagePolicy`.

To grant access to an `ImageRepository` to policies in other namespaces, the owner of the `ImageRepository` has to specify a list of label selectors that match the namespaces of the `ImagePolicy` objects.

```go
type ImageRepositorySpec struct {
	// AccessFrom defines an ACL for allowing cross-namespace references
	// to the ImageRepository object based on the caller's namespace labels.
	// +optional
	AccessFrom *AccessFrom `json:"accessFrom,omitempty"`
}

type AccessFrom struct {
	NamespaceSelectors []NamespaceSelector `json:"namespaceSelector,omitempty"`
}

type NamespaceSelector struct {
	MatchLabels map[string]string `json:"matchLabels,omitempty"`
}
```

When a policy refers to a repository in a different namespace, the controller checks if the policy namespace labels match any of the selectors defined on the ImageRepository object. If the namespace where the policy resides doesn't have labels or the labels don't match the repository ACL, then the controller sets the policy `Ready` status to false and the reason to `AccessDenied`. The access denied error message is set on the `Ready` condition message and logged before the controller rejects the policy.

## Example 

Assuming you have an `apps` namespace where the image pull secrets are. Instead of creating all image automation objects in the `apps` namespace and copy the Git auth secret from `flux-system` into `apps`, you can only define the `ImageRepository` in `apps` and grant access to `ImagePolicies` defined in `flux-system` namespace:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: apps
---
kind: ImageRepository
metadata:
  name: app1
  namespace: apps
spec:
  interval: 5m
  image: docker.io/org/image
  secretRef:
    name: regcred
  accessFrom:
    namespaceSelectors:
      - matchLabels:
          app.kubernetes.io/instance: flux-system
```

The `ImagePolicies` defined in `flux-system` namespace can now reference the  `app1` repository from `apps` namespace:

```yaml
kind: ImagePolicy
metadata:
  name: app1
  namespace: flux-system
spec:
  imageRepositoryRef:
    name: app1
    namespace: apps
  policy:
    semver:
      range: 1.0.x
```

If the owner of an `ImageRepository` wants to grant access to the repo for **all namespaces** in a cluster, this can be done with an empty `matchLabels`:

```yaml
kind: ImageRepository
metadata:
  name: app1
  namespace: apps
spec:
  interval: 5m
  image: docker.io/org/image
  secretRef:
    name: regcred
  accessFrom:
    namespaceSelectors:
      - matchLabels: {}
```

## Feedback & Testing

Please comment on this PR and let us know your thoughts about this feature.

To install the controller with the ACL capability:

1. install the controllers
```sh
flux install --components-extra=image-reflector-controller,image-automation-controller
```
  
2.  apply the CRDs from this branch
```sh
kubectl apply -k https://github.com/fluxcd/image-reflector-controller/config/crd?ref=image-acl
```

3. allow the controller to list namespaces
```sh
cat << EOF | kubectl apply -f -
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: crd-controller-list-ns
rules:
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
  - list
  - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: crd-controller-list-ns
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: crd-controller-list-ns
subjects:
  - kind: ServiceAccount
    name: image-reflector-controller
    namespace: flux-system
EOF
```

4.  deploy image-reflector-controller build of this branch

```
kubectl -n flux-system set image deployment/image-reflector-controller \
manager=ghcr.io/fluxcd/image-reflector-controller:acl-3f77178a
```


squaremo:
Is it possible to gain access to an `ImageRepository` by creating a namespace with the requisite labels? Yes, you would have to know what those are, and yes, you would need the ability to create namespaces -- but this does seem like it's violating the spirit of _granting_ access.

stefanprodan:
> Is it possible to gain access to an ImageRepository by creating a namespace with the requisite labels?

Yes in the same way it's possible to access an app endpoint cross-namespace guarded by a network policy. I should've specify that this proposal is inspired by the Kubernetes NetworkPolicy API.

> Yes, you would have to know what those are, and yes, you would need the ability to create namespaces

Given our tenancy approach showcased here https://github.com/fluxcd/flux2-multi-tenancy this is not possible, a tenant can't create its own namespaces nor it can change the labels of an existing namespace.  In regards to https://github.com/fluxcd/flux2/pull/582, a tenant has access to namespaced objects (due to the role binding) so it can't alter namespaces unless the cluster admins explicitly grants this permission.

squaremo:
> this proposal is inspired by the Kubernetes NetworkPolicy API

Seems to be controversial there, too: https://github.com/kubernetes/kubernetes/issues/88253

There is a decent solution that arose in that issue: let people select namespaces by name. That way, if you care you can at least close that loophole (and if you don't, just select everything). Namespaces have a label with their name, in Kubernetes 1.21.

stefanprodan:
> There is a decent solution that arose in that issue: let people select namespaces by name.

Yes, I don't expect everyone to be on 1.21 but we should document this in the API spec and tell people to use the readonly name label if they upgrade.

controllers/imagepolicy_controller.go
```diff
@@ -107,6 +114,21 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	// check if we are allowed to use the referenced ImageRepository
+	if _, err := r.hasAccessToRepository(ctx, req, pol.Spec.ImageRepositoryRef, repo.Spec.AccessFrom); err != nil {
+		imagev1.SetImagePolicyReadiness(
+			&pol,
+			metav1.ConditionFalse,
+			"AccessDenied",
```
squaremo, line 122: const?

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +306,58 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+	}
+
+	// deny access if the repository has no ACL defined
+	if acl == nil {
+		return false, fmt.Errorf("ImageRepository '%s/%s' can't be accessed due to missing access list",
+			repo.Namespace, repo.Name)
+	}
+
+	// grant access if the repository ACL has no namespace selectors
+	if acl != nil && len(acl.NamespaceSelectors) == 0 {
```
squaremo, line 323: Surely no selectors (rather than an empty selector) means "allow none"?

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +306,58 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+		return true, nil
+	}
+
+	// get the policy namespace labels
+	var policyNamespace v1.Namespace
+	if err := r.Get(ctx, types.NamespacedName{Name: policy.Namespace}, &policyNamespace); err != nil {
+		return false, err
+	}
+	policyLabels := policyNamespace.GetLabels()
+
```
squaremo, line 333: I believe all the following logic is encapsulated in https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelectorAsSelector and https://pkg.go.dev/k8s.io/apimachinery@v0.22.0/pkg/labels#Selector.

controllers/policy_test.go
```diff
@@ -173,7 +174,7 @@ var _ = Describe("ImagePolicy controller", func() {
 		})
 	})
 
-	Context("Using FilterTags", func() {
+	Context("Filers tags", func() {
```
squaremo, line 177:
```suggestion
	Context("Filters tags", func() {
```

controllers/policy_test.go
```diff
@@ -314,4 +315,320 @@ var _ = Describe("ImagePolicy controller", func() {
 			})
 		})
 	})
+
+	Context("Access ImageRepository", func() {
```
squaremo, line 319: Hooray for comprehensive tests! Can the body of the tests be factored out?

squaremo (changes requested):
Presumably much of this can move to fluxcd/pkg/ at some point, since it's pretty general. In the meantime, I think
 - the meaning of no selectors; and,
 - using metav1 for doing the matching
are worth addressing (either with changes or comments here ..)

controllers/imagepolicy_controller.go
stefanprodan: I'll leave it as it is, when this code will be factored out into `fluxcd/pkg` the `AccessDenied` reason will be part of the `apis/meta`.

controllers/policy_test.go
stefanprodan: I'll leave this for when we'll refactor all tests and get rid of Ginko.

controllers/imagepolicy_controller.go
squaremo: OK

controllers/policy_test.go
squaremo: That's a bit of a dodge, but OK if you promise :-)

api/v1beta1/imagerepository_types.go
```diff
@@ -67,6 +67,19 @@ type ImageRepositorySpec struct {
...
 	Suspend bool `json:"suspend,omitempty"`
+
+	// AccessFrom defines an ACL for allowing cross-namespace references
+	// to the ImageRepository object based on the caller's namespace labels.
+	// +optional
+	AccessFrom *AccessFrom `json:"accessFrom,omitempty"`
+}
+
+type AccessFrom struct {
+	NamespaceSelectors []NamespaceSelector `json:"namespaceSelector,omitempty"`
```
squaremo, line 78: The field name and the JSON name don't match. I think the plural is probably what you want:

```yaml
accessFrom:
  namespaceSelectors:
  - matchLabels: {k8s.io/namespace: podinfo-app} # or whatever the official label is
  - matchLabels: {use: app, app: podinfo} # just to make some up ...
```

controllers/imagepolicy_controller.go
squaremo: What was the resolution?

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +307,54 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+	// deny access if the policy namespace has no labels
```
squaremo, line 335: This will deny access even if there's a selector that admits everything.

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +307,54 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+
+	// check if the policy namespace labels match any ACL
+	var allowed bool
+	for _, selector := range acl.NamespaceSelectors {
+		sel, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selector.MatchLabels})
+		if err != nil {
+			return false, err
+		}
+		if sel.Matches(labels.Set(policyLabels)) {
+			allowed = true
```
squaremo, line 349: You can just return `true, nil` here, and return `false, fmt.Errorf(...)` unconditionally after the loop, since matching here or finishing the loop without a match are the only two outcomes.

api/v1beta1/imagerepository_types.go
stefanprodan: The naming is the same as with network policies `namespaceSelector`.

stefanprodan:
@squaremo I've added the Kubernetes readony label to docs, hopefully people will change their selectors once they upgrade to 1.21.

api/v1beta1/imagerepository_types.go
squaremo: Yes, but there it is a matchLabels, not a list of matchLabels. The factoring for network policy is

```yaml
  from:
    - ipBlock:
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - namespaceSelector:
        matchLabels:
          project: myproject
    - podSelector:
        matchLabels:
          role: frontend
```

as in, there's a list of different ways of selecting pods, one kind of which is `NamespaceSelector`, which has a single `matchLabels`.

controllers/imagepolicy_controller.go
stefanprodan: `if acl != nil && acl.NamespaceSelectors == nil`

api/v1beta1/imagerepository_types.go
squaremo: If you want it to be the same as network policy, you should have:

```yaml
allowFrom:
- namespaceSelector:
    matchLabels: {...}
- namespaceSelector:
    matchLabels: {...}
```

stefanprodan: Fixed!

controllers/imagepolicy_controller.go
stefanprodan: Should be fixed by latest commit.

controllers/imagepolicy_controller.go
stefanprodan: Fixed!

docs/spec/v1beta1/imagerepositories.md
```diff
@@ -110,6 +115,50 @@ kubectl create secret generic $SECRET_NAME \
...
+  - namespaceSelector:
+      matchLabels:
+        kubernetes.io/metadata.name: flux-system
+```
+
+**Note** that the `kubernetes.io/metadata.name` is a readonly label added by Kubernetes >= 1.21
+automatically on namespaces. If you're using an older version of Kubernetes, please set labels
+on the namespaces where the `ImagePolicy` are.
+
+The above definition, allows for `ImagePolicy` in the `flux-system` namespace
```
squaremo, line 144: Is it a realistic scenario to have the `ImageRepository` in the namespace `apps`, but the policy objects in `flux-system`? Is that because the likelihood is people will have an automation object in flux-system so it can access the "system" `GitRepository`? (If so, let's remember to rewrite the examples once it's possible to do something different).

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +307,42 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+		sel, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{MatchLabels: selector.MatchLabels})
+		if err != nil {
+			return false, err
+		}
+		if sel.Matches(labels.Set(policyLabels)) {
+			return true, nil
+		}
+	}
+
+	return false, fmt.Errorf("ImageRepository '%s/%s' can't be accessed due to lables mismatch on namespace '%s'",
```
squaremo, line 346:
```suggestion
	return false, fmt.Errorf("ImageRepository '%s/%s' can't be accessed due to labels mismatch on namespace '%s'",
```

controllers/imagepolicy_controller.go
```diff
@@ -284,3 +307,42 @@ func (r *ImagePolicyReconciler) patchStatus(ctx context.Context, req ctrl.Reques
...
+	}
+
+	// deny access if the repository has no ACL defined
+	if acl == nil {
+		return false, fmt.Errorf("ImageRepository '%s/%s' can't be accessed due to missing access list",
+			repo.Namespace, repo.Name)
+	}
+
+	// grant access if the repository ACL has no namespace selectors
+	if acl != nil && acl.NamespaceSelectors == nil {
```
squaremo, line 324: I still think this is odd -- no selectors should mean no access (because none can match). I would just remove this special case.

squaremo (changes requested):
One typo, and the remaining question over what no selectors means.

docs/spec/v1beta1/imagerepositories.md
stefanprodan: Yes it boils down to the secret that `flux bootstrap github --read-write-key` creates that no one wants to copy it to other namespaces.

controllers/imagepolicy_controller.go
stefanprodan: Hmm but `accessFrom: {}` should allow all right?

stefanprodan: Also this should allow all:
```yaml
  accessFrom:
    namespaceSelectors:
      - matchLabels: {}
```

squaremo: That should read as "access from nothing", I would think. Think of it like this: the contents of `accessFrom` specify all the circumstances in which access is allowed. An empty value (nil or empty list) means no circumstances.
In contrast, the contents of `matchLabels` gives all the labels that must match -- if an empty map, there are no requirements (i.e., any set of labels trivially matches).

stefanprodan: Ok so to allow access from all namespaces, we'll only support `matchLabels: {}`?

squaremo: One fiddliness is whether `accessFrom: {}` and `accessFrom: {namespaceSelectors: nil}` mean no access whatsoever, or access only from the same namespace.

stefanprodan: I think I covered the empty ACL (deny) and empty labels (allow) with tests in latest commit.

wavemoran:
I'm building up some new clusters and migrating our workloads from flux v1 to v2. This route makes more logical sense to me as I can now keep a set of apps, repo secrets, and associated `ImagePolicy` and `ImageUpdateAutomation` within the same namespace.

It is also helpful as we restrict engineer teams by namespace. This will help avoid needing access across namespaces or relying on another team to manage flux.

squaremo (approved):
OK cool, looks good to me. I'm glad to have an answer for the people who ask after this! Thank you Stefan 🍇

hiddeco (approved):
LGTM, looking forward to seeing this translated to a `fluxcd/pkg` contribution :-)

PR #164, opened 2021-08-13
Update installation CLI with kubectl

pomverte:
Context:

There is no need to install kustomize cli, it's already embedded in kubectl


hiddeco:
Thank you for your PR 🙏

This is actually done on purpose! The Kustomize version which ships with `kubectl` is not up-to-date with the latest Kustomize releases, which are manifests are designed to follow. This means that while the `kubectl` command may work now, it can break in the future.

pomverte:
👌 

thank you for for answer @hiddeco

PR #165, opened 2021-08-16
Port controller to new pkg/runtime API

squaremo:
- [x] Rewrite conditions code to use API in pkg/runtime/conditions (`MarkTrue`/`...False`/`...Unknown`)
 - [x] Use patch helper rather than own `patchStatus`
 - [x] Exit early without setting status when `.spec.suspend`
 - [x] Events and metrics helpers
 - [ ] Figure out what other things need porting and put them here

Ref: https://github.com/fluxcd/flux2/issues/1601


controllers/imagerepository_controller.go
```diff
@@ -100,23 +103,30 @@ func (r *ImageRepositoryReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		}, patch.WithStatusObservedGeneration{}); err != nil {
+			retErr = kerrors.NewAggregate([]error{retErr, err})
+		}
+	}()
+
 	defer r.recordSuspension(ctx, imageRepo)
 
 	log := logr.FromContext(ctx)
 
 	if imageRepo.Spec.Suspend {
```
hiddeco, line 122: In the source-controller we do not push any updates to the object in case it has been suspended (see: https://github.com/fluxcd/source-controller/blob/reconcilers-dev/controllers/bucket_controller.go#L95-L102).

Making this visual to the end-user should instead be done by including the `Suspend` as a column in the Custom Resource Definition, so that the value on the object is always reflected to table-views, etc.

squaremo: What's the rationale for not accounting for suspend in the status? I guess you can argue that one both ways -- on the one hand, third party libs aren't going to know about `.spec.suspend`, for instance.

squaremo: _goes looking for the previous run of this conversation_

hiddeco: Suspend tells the reconciler to stop looking at the resource and pause all observations, if you reflect this, you are not pausing observations.

squaremo: That's how you've defined it, but why is that better than e.g., recording that it's not ready because it's suspended?

hiddeco: Because continuing to push updates to the object will also continue triggering any watcher that is subscribed to the resource, it will continue bumping the `observedGeneration` on at least the Condition type, and probably some other adverse affects in the machinery area.

It is not just us who uses this approach: https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/machineset_controller.go#L132-L136

squaremo: Cluster API has a pause button so they can implement the `move` command, where they need to stop acting on the resources to be moved because they will be duplicated elsewhere, without upsetting other controllers. I don't think there's an analogue of that requirement for GOTK, but I can probably come up with other scenarios where that behaviour 
 is useful, specifically the "without upsetting other controllers" part.

squaremo: In other words, I'll remove that code if it's part of the GOTK contract.

stefanprodan:
ACLs have moved to fluxcd/pkg:
- Type here: https://github.com/fluxcd/pkg/blob/main/apis/acl/acl_types.go
- Condition here: https://github.com/fluxcd/pkg/blob/main/apis/acl/conditions.go
- Runtime helper here: https://github.com/fluxcd/pkg/blob/main/runtime/acl/acl.go
- Tests here: https://github.com/fluxcd/pkg/blob/main/runtime/acl/acl_test.go

I also think this PR should include test refactoring, replacing Ginko with https://github.com/fluxcd/pkg/blob/main/runtime/testenv/testenv.go

squaremo:
> I also think this PR should include test refactoring

No, I'm not going to push the boat out further and do that here. I don't think it's a matter of compliance.

squaremo:
> ACLs have moved to fluxcd/pkg:

OK -- I would expect that to be its own change.

controllers/imagepolicy_controller.go
```diff
@@ -172,49 +178,38 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		meta.ReadyCondition,
+		meta.SucceededReason,
 		msg,
 	)
-
-	if err := r.patchStatus(ctx, req, pol.Status); err != nil {
-		return ctrl.Result{}, err
-	}
-	r.event(ctx, pol, events.EventSeverityInfo, msg)
+	r.Event(ctx, &pol, events.EventSeverityInfo, EventReasonRefFound, msg)
```
stefanprodan, line 212: Users are reporting this is spammy (see https://github.com/fluxcd/notification-controller/issues/234#issuecomment-904200985). I think we should send an event only if a new image tag has been found. The same way source-controller/helm-controller don't send events on no-op reconciliations (same git revision, chart revision).

squaremo: Seems reasonable; let's do it in another PR though, it's not trivial and I want to strictly limit the scope of this one.

hiddeco (approved):
Sorry it took my rather long to properly review this.

Based on the comments about wanting to address certain things in (a) separate PR(s), this seems to aligned with what the new helpers have to offer.

One personal preference I would like to make a note of is that some of the conditions that now result in `Ready==False` with a descriptive reasons may be candidates for a dedicated negative polarity condition. Reason for this is that it may help consumers that e.g. try to build an UI around the API object to make things visible to their end users, as Condition Types are part of the API (and can thus be used to (conditionally) render things), while the reasons are arbitrary.

In any case, of to a great start, and thank you @squaremo :1st_place_medal:

PR #169, opened 2021-09-14
Add setup-envtest to makefile & gh build action

darkowlzz:
This removes installing kubebuilder binaries and uses setup-envtest to
install all the necessary envtest helper binaries.

⚠️ target branch of the PR is `reconcilers-dev`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@ IMG ?= fluxcd/image-reflector-controller
 # Produce CRDs that work back to Kubernetes 1.16
 CRD_OPTIONS ?= crd:crdVersions=v1
 
+ENVTEST_BIN_VERSION?=1.22.0
+KUBEBUILDER_ASSETS?=$(shell $(SETUP_ENVTEST) use -i $(ENVTEST_BIN_VERSION) -p path)
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,6 +3,9 @@ IMG ?= fluxcd/image-reflector-controller
 # Produce CRDs that work back to Kubernetes 1.16
 CRD_OPTIONS ?= crd:crdVersions=v1
 
+ENVTEST_BIN_VERSION?=1.22.0
+KUBEBUILDER_ASSETS?=$(shell $(SETUP_ENVTEST) use -i $(ENVTEST_BIN_VERSION) -p path)
```
squaremo, line 7: I couldn't remember if `?=` expanded once or at every mention, so I looked it up: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC59 (answer: yes, it's expanded at every mention). This is usually a bit dangerous. On the other hand, it wouldn't work to use a simple expansion (`:=`) because it relies on SETUP_ENVTEST, which is only set in the rule for `setup_envtest`. I can't think of a nicer way to do this though, can you?

squaremo (approved):
Yup, this is neater -- thanks @darkowlzz 🥝

Makefile
darkowlzz: No, I've only seen this pattern everywhere. I'm not aware of anything nicer.

darkowlzz:
I've changed the target branch to `reconcilers-dev` branch because that's were the new development is taking place and I would like this change there to help with running the tests.

PR #170, opened 2021-09-17
Testenv test refactor

darkowlzz:
Introduces testenv based tests, ~~keeping the ginkgo test suite in legacy_suite_test.go~~.
Refactor all the tests to use testenv.

It uses fluxcd/pkg/runtime from https://github.com/fluxcd/pkg/pull/148 to be able to
run badger in the test.

More details about the changes are in the commit body.

:warning:  target branch of the PR is `reconcilers-dev`.

PR #172, opened 2021-09-20
policy: Handle failure due to invalid semver range

### LINKED ISSUE #171: image-reflector-controller segfault on bad semver range
### Describe the bug

When this ImagePolicy was deployed, the image-reflector-controller started crashlooping with the error that follows. Taking the second wildcard out of the `range:` value stopped the segfault crashloop.

```
kind: ImagePolicy
metadata:
  name: our-policy
  namespace: default 
spec:
  imageRepositoryRef:
    name: favorite-repo
  policy:
    semver:
      range: "*-*"
```

Log output:
```
2021-09-17 13:54:43.239 PDTpanic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1657475] goroutine 334 [running]: github.com/fluxcd/image-reflector-controller/internal/policy.(*SemVer).Latest(0x0, 0xc000a5b8c0, 0x6, 0x6, 0x6, 0x6, 0x0, 0x0) /workspace/internal/policy/semver.go:55 +0x135 github.com/fluxcd/image-reflector-controller/controllers.(*ImagePolicyReconciler).Reconcile(0xc000458cd0, 0x1c00e18, 0xc001143c50, 0xc000a86c40, 0x7, 0xc0009aad80, 0x13, 0xc001143c00, 0x0, 0x0, ...) /workspace/controllers/imagepolicy_controller.go:146 +0x125a sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000a237c0, 0x1c00d70, 0xc00092e380, 0x182d140, 0xc000aaefc0) /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:298 +0x30d sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc000a237c0, 0x1c00d70, 0xc00092e380, 0x442700) /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:253 +0x205 sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2(0xc000c51420, 0xc000a237c0, 0x1c00d70, 0xc00092e380) /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:214 +0x6b created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2 /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.5/pkg/internal/controller/controller.go:210 +0x425
```

### Steps to reproduce

1. deploy an ImagePolicy with semver range "*-*"
2. image-reflector-controller crashes

### Expected behavior

No crash.

### Screenshots and recordings

_No response_

### OS / Distro

GKE 1.20

### Flux version

0.17.0

### Flux check

N/A

### Git provider

_No response_

### Container Registry provider

_No response_

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

darkowlzz:
This adds error check after creating a Policer.
For alphabetical and numerical policies, the k8s API validates the
input data. But for semver policy, there aren't predefined valid
values.

Adds a test to verify the fix.

Fixes #171 

Reference to the same fix and tests against `reconcilers-dev` branch
after [testenv test refactor](https://github.com/fluxcd/image-reflector-controller/pull/170) :  https://github.com/darkowlzz/image-reflector-controller/compare/testenv-test-refactor...darkowlzz:invalid-policy-semver-range?expand=1, table test with new test case.


stefanprodan (approved):
LGTM 

Thanks @darkowlzz 🥇

controllers/imagepolicy_controller.go
```diff
@@ -148,6 +148,20 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	var err error
```
squaremo, line 149: Is this shadowed right away? If I'm reading this right, probably worth just removing this declaration as part of this change, for clarity. And using a differently-named var to assign to in the loop a few lines below.

squaremo: OK I think I see what's going on. The expectation is that the value returned from `PolicerFromSpec` _either_ has non-nil `policer` or non-nil `err`; so if there's an error, the block which accesses policer won't run, and the conditional following that will catch errors from parsing the policy _or_ from applying the policy. All else being correct, this would work fine.

However, `policer` is typed as `Policer`, whereas the return value is `*SemVer`; this type checks, but `nil(Policer) != nil(*SemVer)`, so the block conditional on `policer != nil` runs after all.  Am I right?

squaremo: > However, policer is typed as Policer, whereas the return value is *SemVer; this type checks, but nil(Policer) != nil(*SemVer), so the block conditional on policer != nil runs after all. Am I right?

Yes I think I am. If I remove the early exit you added, and put `fmt.Printf("%s\n", policer)` _inside_ the block conditional on `policer != nil`, I can see 

    [DEBUG] %!s(*policy.SemVer=<nil>)

printed in the output of

    go test -v ./controllers/...

(and a rightly failing test). Though the fix here already avoids it crashing, I think it'd be better to fix that underlying problem as well.

squaremo (changes requested):

controllers/imagepolicy_controller.go
darkowlzz: Yeah, that's seems to be what was happening. And later on when `policer.Latest()` is called, it panicked. An interface holding a nil value not equal to nil. The policer nil check will only protect against the situation where the choice is something other than the known choices, but I don't see that'll really ever happen. We should be able to remove the policer check.

squaremo: I think it would pay to make `PolicerFromSpec` return Policer(nil) values when it gets an error, otherwise this will remain as a trap for future changes.

darkowlzz: Yeah, adding that. Thanks for digging into it.

darkowlzz:
Adding a test for it as well, please don't merge it.

darkowlzz:
Added a test for `PolicyFromSpec()` in the factory tests to ensure that it's returns a nil checkable Policer. This is ready now :slightly_smiling_face:

controllers/imagepolicy_controller.go
squaremo, line 149: This declaration could still do with removing, it's redundant. I think the shadowing only escapes the linter because policer is a new variable.
I'd also like to avoid the reuse of `err` afterward. It made some sense before (it might be the error returned by `PolicerFromSpec`), but now it's confusing.

squaremo (approved):
I leave my last suggestion at your disposal. Thanks Sunny!

controllers/imagepolicy_controller.go
darkowlzz: I've just move this declaration just above the block that fetches the latest tag.

darkowlzz: I'll just move this error before the block that fetches the latest tag with a different name.

darkowlzz: Changed my mind, just removing the first declaration :sweat_smile:

darkowlzz:
This is ready from my side.

PR #147, opened 2021-06-01
Log in to AWS ECR automatically using the cluster-provided credentials.

MartinEmrich:
As discussed in fluxcd/image-reflector-controller#139

This mostly restores the flux v1 functionality (for me)

if the environment variable USE_ECR is set, logs in to the AWS Elastic
Container Registry in the image URL by fetching a temporary token from
AWS.

Issue #139, opened 2021-04-28
Authentication failure with ECR


davisford:
The image reflector controller log is showing this:

```
{"level":"error","ts":"2021-04-27T14:55:40.005Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"giftcard","namespace":"flux-system","error":"GET https://XXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/v2/redacted/redacted/tags/list?n=1000: unexpected status code 401 Unauthorized: Not Authorized\n"}
```

Which is bizarre because the node in EKS has this policy attached to it:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ecr:GetAuthorizationToken",
                "ecr:BatchCheckLayerAvailability",
                "ecr:GetDownloadUrlForLayer",
                "ecr:GetRepositoryPolicy",
                "ecr:DescribeRepositories",
                "ecr:ListImages",
                "ecr:DescribeImages",
                "ecr:BatchGetImage",
                "ecr:GetLifecyclePolicy",
                "ecr:GetLifecyclePolicyPreview",
                "ecr:ListTagsForResource",
                "ecr:DescribeImageScanFindings"
            ],
            "Resource": "*"
        }
    ]
}
```

...and if I specify an image tag for the same repository, Kubernetes has no problem at all pulling the image.

I understand there is a mechanism you recommend to setup a secret + daemon set to get the credentials, but I'm puzzled as to why it can't just pull the image when the node itself has permissions to do so, and Kubernetes pulls the images just fine.

relu:
Hello @davisford, thanks for submitting the issue.

I think your question is a valid one and I hope that you'll find my answer useful.
[...]

PR #174, opened 2021-09-30
Get credentials from AWS ECR when needed

squaremo:
A rebase of #147 (to avoid force pushing over a main branch).

> As discussed in fluxcd/image-reflector-controller#139
> 
> If the flag `--user-aws-ecr` is set, the controller will log in to AWS
> Elastic Container Registry by fetching a temporary token from AWS
> (when a given image is hosted in ECR).

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -21,11 +21,13 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -44,6 +46,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
 
 	"github.com/fluxcd/pkg/apis/meta"
 	"github.com/fluxcd/pkg/runtime/events"
@@ -74,6 +80,7 @@ type ImageRepositoryReconciler struct {
 		DatabaseWriter
 		DatabaseReader
 	}
+	UseAwsEcr bool
 }
 
 type ImageRepositoryReconcilerOptions struct {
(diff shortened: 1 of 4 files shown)


squaremo:
I would love to see tests for this, but mocking AWS services is no small task. Perhaps something like https://github.com/localstack/localstack could help.

main.go
```diff
@@ -79,6 +80,8 @@ func main() {
 	flag.StringVar(&storagePath, "storage-path", "/data", "Where to store the persistent database of image metadata")
 	flag.Int64Var(&storageValueLogFileSize, "storage-value-log-file-size", 1<<28, "Set the database's memory mapped value log file size in bytes. Effective memory usage is about two times this size.")
 	flag.IntVar(&concurrent, "concurrent", 4, "The number of concurrent resource reconciles.")
+	flag.BoolVar(&useAwsEcr, "use-aws-ecr", false, "Log in to AWS Elastic Container Registry with IAM")
```
stefanprodan, line 83:
```suggestion
	flag.BoolVar(&useAwsEcr, "enable-aws-iam", false, "Log in to AWS Elastic Container Registry with IAM")
```

Sugested by Hidde in the original PR.

controllers/imagerepository_controller.go
```diff
@@ -217,6 +265,28 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1
...
+					meta.ReconciliationFailedReason,
+					err.Error(),
+				)
+				return err
+			}
+
+			auth := authn.FromConfig(authConfig)
+			options = append(options, remote.WithAuth(auth))
+		} else {
+			logr.FromContext(ctx).Info("AWS ECR authentication is not enabled, to enable, set USE_ECR environment variable")
```
stefanprodan, line 286:
```suggestion
			logr.FromContext(ctx).Info("AWS ECR authentication is not enabled, to enable, set --enable-aws-iam")
```

stefanprodan:
@squaremo does this work with IRSA or with node IAM bindings or with both?

squaremo:
> does this work with IRSA or with node IAM bindings or with both?

From the explanation in https://github.com/fluxcd/image-reflector-controller/issues/139#issuecomment-849421642 I would expect it works with the node IAM bindings that EKS does automatically, as well as IRSA (IAM roles for service accounts) and possibly other mechanisms, since "one can grant additional IAM permissions via a multitude of mechanics (IAM node role, IAM service account, or even configuring access keys manually via environment variables), and all official AWS SDKs will automatically try each until it gets a valid identity". @MartinEmrich Are you able to confirm that this is the correct expectation?

MartinEmrich:
@squaremo @stefanprodan

It works with the IAM instance profile that each EKS node has to have. 
I guess AWS puts the ecr credentials helper (https://github.com/awslabs/amazon-ecr-credential-helper) in their provided AMIs, or they run a patched kubelet which regularly gets an ECR login token.
See also: https://docs.aws.amazon.com/AmazonECR/latest/userguide/ECR_on_EKS.html
The authenticated entity here is the Node/Kubelet itself, not the user workload.

Container image pulling should have no relationship to IAM service accounts/IRSA (after all, service accounts are assigned to pods, which run an already pulled image).

IRSA (or, putting access keys somewhere in your workloads, which is always bad and unneccesary), operates within your workloads (Pods), so are on a completely different level.

(Disclaimer: I a just a dude with a google, I have no hidden insights in how AWS does stuff internally.)

squaremo:
Regarding the name of the flag: I don't really have a strong opinion on `use` vs `enable` vs `with`; consistency can guide us there. But I would like it to be precise and descriptive, so I would go for `--use-aws-iam-for-ecr` or `--aws-use-iam-for-ecr` (I'm undecided on how I would want the flags to sort, whether by `aws` or by `enable` or `use`; but again, consistency should be the guide).

MartinEmrich:
As for the flag, I would like to urge you just one final time to make it enabled by the default, to follow the Principle Of Least Surprise. Again, this feature has no security implications, drawbacks or unexpected behaviour. (IMHO, even a flag to disable it makes no sense, as it does not protect anything from anything)

squaremo:
This is my understanding:

 - Flux v1 uses [ecrClient.GetAuthorizationToken](https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#ECR.GetAuthorizationToken). This will work if the pod has permissions to read ECR in the account given.
   - A controller pod in EKS will generally have read permissions to ECR, granted through the [node IAM role](https://docs.aws.amazon.com/eks/latest/userguide/create-node-role.html) (mysteriously, the mechanism whereby pods get the node permissions is not mentioned there, but it is implied [elsewhere](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html#restrict-ec2-credential-access))
   - Another possible mechanism for granting permissions _in EKS_ is using [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) to give the controller service account the appropriate IAM role; it's possible to block the node permissions being granted to pods, in which case this would become necessary.
   - If you create an EKS cluster [with eksctl or by following the getting started guide](https://docs.aws.amazon.com/AmazonECR/latest/userguide/ECR_on_EKS.html) then the cluster nodes are given permission to read ECR
 - Flux v1 will only use this facility if you specifically allow it to, or if it detects (via the metadata service) that it's running in EC2 (which [has its own problems](https://github.com/fluxcd/flux/issues/3015)).
 - At present, Flux v2 does not try to get authentication credentials for ECR.

This PR unconditionally asks ECR for credentials (given the enable flag, or if it were always enabled). That has the following effect:

| Controller environment | Effect |
|-----|-----|
| EKS | By default, all `ImageRepository` objects can read ECR for the account under which EKS cluster is run |
| AWS in general | All `ImageRepository` objects can read ECR according to assigned IAM permissions |
| Outside EC2 | For ECR images, `ecr.GetAuthorizationToken` will fail and the `ImageRepository` object will be marked unready (whether or not another secret was supplied)

So I think there are two reasons to guard this feature with a flag:

 - if you are not running in AWS, but you are using ECR for some reason, you don't want it to fail unconditionally (this can be fixed in the code by only trying `ecr.GetAuthorizationToken` when there's no secret referenced)
 - Usually, every pod can read everything in (the cluster account in) ECR, and having `ImageRepository` objects work differently would be strange; however, it's [possible to remove that global permission for pods](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html#restrict-ec2-credential-access), so it ought to be possible to remove it for `ImageRepository` objects.

stefanprodan:
> if you are not running in AWS, but you are using ECR for some reason, you don't want it to fail unconditionally (this can be fixed in the code by only trying ecr.GetAuthorizationToken when there's no secret referenced)

I think this the way to go, especially since many orgs don't use the same account for ECR and EKS. We can't assume that credentials present on nodes are for the registry that hosts the app images.

PR #189, opened 2021-10-27
Set default observedGeneration to -1

tomhuang12:
This sets the status.observedGeneration field to -1 by default. This is to address issue outlined by https://github.com/fluxcd/flux2/issues/2007. 

Signed-off-by: Tom Huang <tom.huang@weave.works>


makkes (approved):
Thanks @tomhuang12

hiddeco (approved):

PR #191, opened 2021-10-29
Add various instructions on development documentation

tomhuang12:
This is to address https://github.com/fluxcd/image-reflector-controller/issues/182.


DEVELOPMENT.md
```diff
@@ -16,3 +23,30 @@ You can run the unit tests by simply doing
...
+```bash
+make run
+```
+
+## How to generate and update CRDs API reference documentation
+
+You can update CRDs API reference doc by
+
+```bash
+make gen-crd-api-reference-docs
```
relu, line 51: I think this should be `make api-docs`, `gen-crd-api-reference-docs` will only install the tool.

DEVELOPMENT.md
```diff
@@ -16,3 +23,30 @@ You can run the unit tests by simply doing
...
+```
```
relu, line 52: Maybe also add a note here somewhere about referring to the Makefile. All the build targets have comments about what they do.

DEVELOPMENT.md
```diff
@@ -16,3 +23,30 @@ You can run the unit tests by simply doing
 ```bash
 make test
```
relu, line 24: Also a note here about running this prior to committing any changes, especially in the situation when API changes are made.

DEVELOPMENT.md
```diff
@@ -4,6 +4,13 @@
 > to find out about how to contribute to Flux and how to interact with the
 > Flux Development team.
 
+## Installing required dependencies
+There are a number of dependencies required to be able to run image-reflector-controller and its test-suite locally. 
+* [Install Go](https://golang.org/doc/install)
+* [Install Kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+* [Install Docker](https://docs.docker.com/engine/install/)
+* [Install Kubebuilder](https://book.kubebuilder.io/quick-start.html)
```
relu, line 12: I don't think installing Kubebuilder is mandatory. It's probably useful only for initial scaffolding. Maybe have it as optional?

relu (changes requested):
Great work, @tomhuang12! :raised_hands:
I left a few comments, hope it helps.

tomhuang12:
@relu Thanks for the review. Please see updated changes.

DEVELOPMENT.md
```diff
@@ -4,15 +4,51 @@
...
+
 ## How to run the test suite
 
 Prerequisites:
 * go >= 1.16
-* kubebuilder >= 2.3
 * kustomize >= 3.1
 
-You can run the unit tests by simply doing
+Prior committing any changes, especially when API changes are made, please run unit tests. You can run them by simply doing
```
relu, line 20: Maybe keep the original and just add the note after the `make test` code block.

Something like this, also explains the reason behind doing that:
> Since this will also trigger generating some files such as manifests, it is advised to run this prior to committing your changes, especially when making API changes.

dholbach:
Can you rebase your PR?

tomhuang12:
It is rebased now. Thanks

kingdonb (changes requested):
LGTM 👍 but the rebase is a bit off

@tomhuang12 Your PR is still based off of: 2d05314 while head is currently: `62c06ea`

I think you need to fast-forward advance your local `main` branch before rebasing, could you take another look at this?

(I do not have triage here, nor write access, so my review does not count)

tomhuang12:
I messed up the rebase. I am going to re-open a PR.

PR #193, opened 2021-10-31
Elaborate on how to enable autologin for ECR

kingdonb:
Fixes fluxcd/website#601

There is not enough context in the documentation for `--aws-autologin-for-ecr` to guarantee that a new user who finds this section of the doc will be able to figure out where or how to add this flag to the `image-reflector-controller`.

Provide an example `patches` directive and be sure to say more explicitly which controller accepts this flag. (It is not at all obvious which controller owns the documentation at this URL, even though it has been pulled from this repo, there is no breadcrumb at all or anything else which says "image reflector controller")

https://fluxcd.io/docs/components/image/imagerepositories/#ecr-and-eks

### DIFF
--- a/docs/spec/v1beta1/imagerepositories.md
+++ b/docs/spec/v1beta1/imagerepositories.md
@@ -61,7 +61,7 @@ type ImageRepositorySpec struct {
 ```
 
 The `Suspend` field can be set to `true` to stop the controller scanning the image repository
-specified; remove the field value or set to `false` to resume scanning.
+specified; remove the field value or set it to `false` to resume scanning.
 
 ### Authentication
 
@@ -98,7 +98,19 @@ patches:
         value: --aws-autologin-for-ecr
 ```
 
-Alternatively, the advice under "Other platforms" below will also work for ECR.
+Alternatively, the advice under ["Other platforms"](https://github.com/fluxcd/image-reflector-controller/blob/main/docs/spec/v1beta1/imagerepositories.md#other-platforms) below will also work for ECR.
+
+> You need to upgrade to Flux version 2 release [v0.19](https://github.com/fluxcd/flux2/releases/tag/v0.19.0) that contains the image-reflector-controller release [v0.13.0](https://github.com/fluxcd/image-reflector-controller/blob/main/CHANGELOG.md#0130).
+
+> [**Release date**: 2021-10-19](https://github.com/fluxcd/image-reflector-controller/blob/main/CHANGELOG.md#0130)
+>
+> This prerelease adds experimental support for automatically getting
+credentials from AWS when scanning an image in [Elastic Container Registry
+(ECR)](https://docs.aws.amazon.com/AmazonECR/latest/userguide/what-is-ecr.html).
+>
+> Improvements:
+> * Get credentials from AWS ECR when needed
+>  [#174](https://github.com/fluxcd/image-reflector-controller/pull/174)
 
 #### Other platforms
 
@@ -209,7 +221,7 @@ type ImageRepositoryStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
-	// CannonicalName is the name of the image repository with all the
+	// CanonicalName is the name of the image repository with all the
 	// implied bits made explicit; e.g., `docker.io/library/alpine`
 	// rather than `alpine`.
 	// +optional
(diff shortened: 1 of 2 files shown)


kingdonb:
I incorporated some additional feedback from @lloydchang based on kingdonb/image-reflector-controller#1

This review also spotted and fixed some typos, not directly in the path of this change but in adjacent sections, which I think are close enough to include here. Thanks!

docs/spec/v1beta1/imagerepositories.md
```diff
@@ -190,7 +221,7 @@ type ImageRepositoryStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
-	// CannonicalName is the name of the image repository with all the
+	// CanonicalName is the name of the image repository with all the
```
kingdonb, line 224:
```suggestion
	// CanonicalImageName is the name of the image repository with all the
```

This was corrected wrong, it's `CanonicalImageName` per the line below the doc string

lloydchang (approved):

relu (approved):
Looks pretty clear to me :100:

docs/spec/v1beta1/imagerepositories.md
```diff
@@ -77,9 +77,40 @@ For a publicly accessible image repository, you will not need to provide a `secr
...
+      namespace: flux-system
+    patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --aws-autologin-for-ecr
+```
+
+Alternatively, the advice under [Other platforms][other platforms] below will also work for ECR.
+
+> You need to upgrade to Flux version 2 release [v0.19][Flux v0.19.0] that contains the image-reflector-controller release [v0.13.0][image-reflector-controller v0.13.0].
```
stefanprodan, line 103: API specs usually don't contain changelogs, what's the reason we add these here?

squaremo:
> (It is not at all obvious which controller owns the documentation at this URL)

That's because it's API documentation -- no controller owns this documentation. I hesitated to refer to the flag here at all, because advice on how to run the controller really belongs elsewhere (but an elsewhere that doesn't exist, at present).

squaremo:
Please keep the commit message subject to 50 characters -- see https://github.com/fluxcd/.github/blob/main/CONTRIBUTING.md#format-of-the-commit-message

stefanprodan:
I think the patch example should be in the [guide](https://fluxcd.io/docs/guides/image-update/#using-native-aws-ecr-auto-login) not here in the API docs.

docs/spec/v1beta1/imagerepositories.md
kingdonb: As an experimental feature, I wanted to call out the release version that these docs applied to so that it is clear enough what needs to be updated when changes inevitably happen. (Also, it was a suggestion from @lloydchang)

kingdonb:
I'll implement these changes and come back, thank you

docs/spec/v1beta1/imagerepositories.md
```diff
@@ -190,7 +221,7 @@ type ImageRepositoryStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
-	// CannonicalName is the name of the image repository with all the
+	// CanonicalImageName is the name of the image repository with all the
```
darkowlzz, line 224: This section is taken from https://github.com/fluxcd/image-reflector-controller/blob/main/api/v1beta1/imagerepository_types.go#L99 which also has this incorrect name. Can you update that as well and run `make manifests` to generate the same in the CRD spec in `config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml`?

solms:
Not sure how much value this adds, but just to let you know that I am a (new) Flux user who spent _way_ too much time trying to figure out how to use the `--aws-autologin-for-ecr`, and this PR finally cleared it up and sorted me out :)

So +1 to new users being lost in this section and thanks @kingdonb for putting this out there!

docs/spec/v1beta1/imagerepositories.md
kingdonb: It looks like this has been taken care of. I found another CannonicalName in the docs and updated it there 👍

relu (changes requested):

kingdonb:
I am going to close this and reopen based on the current WIP in #702

There are still maybe some valuable additions in here that have not been covered there already. For the most part, this PR is superseded by that one.

PR #194, opened 2021-11-05
Get credentials from GCP/Azure when needed

### LINKED ISSUE #179: GKE/GCR "autologin"
https://github.com/fluxcd/image-reflector-controller/pull/174 adds a flag to the controller that lets it use the AWS client to get credentials for ECR; the argument is that if you're running in EKS (and possibly elsewhere in AWS), all pods will by default have read access to your account in ECR, so you gain a lot in convenience without giving anything away by assuming all `ImageRepository` objects should have access to your ECR account too.

A similar situation exists in GKE with GCR, so it may be similarly convenient (and justifiable) to use a GCP client to get credentials for images in GCR.

Related: https://github.com/fluxcd/image-reflector-controller/issues/11

### LINKED ISSUE #180: Azure/ACR "autologin"
https://github.com/fluxcd/image-reflector-controller/pull/174 adds a flag to the controller that lets it use the AWS client to get credentials for ECR; the argument is that if you're running in EKS (and possibly elsewhere in AWS), all pods will by default have read access to your account in ECR, so you gain a lot in convenience without giving anything away by assuming all `ImageRepository` objects should have access to your ECR account too.

A similar situation exists in Azure (or AKS) with ACR, so it may be similarly convenient (and justifiable) to use a platform-specific client or mechanism to get credentials for images in ACR.

Related: https://github.com/fluxcd/image-reflector-controller/issues/11

SomtochiAma:
Closes #180 
Closes #179 

This pull request adds two flags (`gcp-autologin-for-gcr` and `azure-autologin-for-acr`) for autologin on GCP and Azure to their container registries.

This has been tested on GKE clusters with workload identities enabled and default service account.
This has been tested on an AKS w Managed Identities.

Further refactoring could be done to move implementation details of various cloud providers into a dedicated package.

### DIFF
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -27,6 +27,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      # azure-sdk-for-go requires 1.16 or greater
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.x'
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v1
         with:
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -46,6 +47,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
@@ -56,6 +60,7 @@ import (
 	"github.com/fluxcd/pkg/runtime/predicates"
 
 	imagev1 "github.com/fluxcd/image-reflector-controller/api/v1beta1"
+	"github.com/fluxcd/image-reflector-controller/internal/azure"
 )
 
 // These are intended to match the keys used in e.g.,
@@ -80,7 +85,9 @@ type ImageRepositoryReconciler struct {
 		DatabaseReader
 	}
 
-	AwsAutoLogin bool // automatically attempt to get credentials for images in ECR
+	AwsAutoLogin   bool // automatically attempt to get credentials for images in ECR
+	GcpAutoLogin   bool // automatically attempt to get credentials for images in GCP
+	AzureAutoLogin bool // automatically attempt to get credentials for images in ACR
 }
 
 type ImageRepositoryReconcilerOptions struct {
@@ -89,6 +96,12 @@ type ImageRepositoryReconcilerOptions struct {
 
 type dockerConfig struct {
 	Auths map[string]authn.AuthConfig
+}
+
+type gceToken struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int    `json:"expires_in"`
+	TokenType   string `json:"token_type"`
 }
 
 // +kubebuilder:rbac:groups=image.toolkit.fluxcd.io,resources=imagerepositories,verbs=get;list;watch;create;update;patch;delete
@@ -246,6 +259,53 @@ func getAwsECRLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error
 	return authConfig, nil
 }
 
+// getGCRLoginAuth obtains authentication for the image by
+// getting a token from the metadata API on GCP. This assumes that
+// the pod has right to pull the image which would be the case if it
+// is hosted on GCP. It works with both service account and workload identity
+// enabled clusters.
+func getGCRLoginAuth(ctx context.Context) (authn.AuthConfig, error) {
+	var authConfig authn.AuthConfig
+	const gcpDefaultTokenURL = "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, gcpDefaultTokenURL, nil)
+	if err != nil {
+		return authConfig, err
+	}
+
+	request.Header.Add("Metadata-Flavor", "Google")
+
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		return authConfig, err
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return authConfig, fmt.Errorf("unexpected status from metadata service: %s", response.Status)
+	}
+
+	var accessToken gceToken
+	decoder := json.NewDecoder(response.Body)
+	if err := decoder.Decode(&accessToken); err != nil {
+		return authConfig, err
+	}
+
+	if _, err := io.Copy(io.Discard, response.Body); err != nil {
+		return authConfig, err
+	}
+
+	if err := response.Body.Close(); err != nil {
+		return authConfig, err
+	}
+
+	authConfig = authn.AuthConfig{
+		Username: "oauth2accesstoken",
+		Password: accessToken.AccessToken,
+	}
+	return authConfig, nil
+}
+
 func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1.ImageRepository, ref name.Reference) error {
 	timeout := imageRepo.GetTimeout()
 	ctx, cancel := context.WithTimeout(ctx, timeout)
@@ -296,6 +356,46 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1
 			options = append(options, remote.WithAuth(auth))
 		} else {
 			ctrl.LoggerFrom(ctx).Info("No image credentials secret referenced, and ECR authentication is not enabled. To enable, set the controller flag --aws-autologin-for-ecr")
+		}
+	} else if hostIsGoogleContainerRegistry(ref.Context().RegistryStr()) {
+		if r.GcpAutoLogin {
+			ctrl.LoggerFrom(ctx).Info("Logging in to GCP GCR for " + imageRepo.Spec.Image)
+			authConfig, err := getGCRLoginAuth(ctx)
+			if err != nil {
+				ctrl.LoggerFrom(ctx).Info("error logging into GCP " + err.Error())
+				imagev1.SetImageRepositoryReadiness(
+					imageRepo,
+					metav1.ConditionFalse,
+					meta.ReconciliationFailedReason,
+					err.Error(),
+				)
+				return err
+			}
+
+			auth := authn.FromConfig(authConfig)
+			options = append(options, remote.WithAuth(auth))
+		} else {
+			ctrl.LoggerFrom(ctx).Info("No image credentials secret referenced, and GCR authentication is not enabled. To enable, set the controller flag --gcp-autologin-for-gcr")
+		}
+	} else if hostIsAzureContainerRegistry(ref.Context().RegistryStr()) {
+		if r.AzureAutoLogin {
+			ctrl.LoggerFrom(ctx).Info("Logging in to Azure ACR for " + imageRepo.Spec.Image)
+			authConfig, err := getAzureLoginAuth(ctx, ref)
+			if err != nil {
+				ctrl.LoggerFrom(ctx).Info("error logging into ACR " + err.Error())
+				imagev1.SetImageRepositoryReadiness(
+					imageRepo,
+					metav1.ConditionFalse,
+					meta.ReconciliationFailedReason,
+					err.Error(),
+				)
+				return err
+			}
+
+			auth := authn.FromConfig(authConfig)
+			options = append(options, remote.WithAuth(auth))
+		} else {
+			ctrl.LoggerFrom(ctx).Info("No image credentials secret referenced, and ACR authentication is not enabled. To enable, set the controller flag --azure-autologin-for-acr")
 		}
 	}
 
@@ -590,4 +690,48 @@ func getURLHost(urlStr string) (string, error) {
 	}
 
 	return u.Host, nil
+}
+
+// getAzureLoginAuth returns authentication for ACR. The details needed for authentication
+// are gotten from environment variable so there is not need to mount a host path.
+func getAzureLoginAuth(ctx context.Context, ref name.Reference) (authn.AuthConfig, error) {
+	var authConfig authn.AuthConfig
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return authConfig, err
+	}
+	armToken, err := cred.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{string(arm.AzurePublicCloud)},
+	})
+	if err != nil {
+		return authConfig, err
+	}
+
+	ex := azure.NewExchanger(ref.Context().RegistryStr())
+	accessToken, err := ex.ExchangeACRAccessToken(string(armToken.Token))
+	if err != nil {
+		return authConfig, fmt.Errorf("error exchanging token: %w", err)
+	}
+
+	return authn.AuthConfig{
+		// this is the acr username used by Azure
+		// See documentation: https://docs.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token
+		Username: "00000000-0000-0000-0000-000000000000",
+		Password: accessToken,
+	}, nil
+}
+
+// List from https://github.com/kubernetes/kubernetes/blob/v1.23.1/pkg/credentialprovider/azure/azure_credentials.go#L55
+func hostIsAzureContainerRegistry(host string) bool {
+	for _, v := range []string{".azurecr.io", ".azurecr.cn", ".azurecr.de", ".azurecr.us"} {
+		if strings.HasSuffix(host, v) {
+			return true
+		}
+	}
+	return false
+}
+
+func hostIsGoogleContainerRegistry(host string) bool {
+	return host == "gcr.io" || strings.HasSuffix(host, ".gcr.io") || strings.HasSuffix(host, "-docker.pkg.dev")
 }
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,8 @@ go 1.17
 replace github.com/fluxcd/image-reflector-controller/api => ./api
 
 require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.20.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.12.0
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/aws/aws-sdk-go v1.42.9
 	github.com/dgraph-io/badger/v3 v3.2103.2
@@ -23,6 +25,7 @@ require (
 )
 
 require (
+	github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.1 // indirect
 	cloud.google.com/go v0.97.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash v1.1.0 // indirect
@@ -61,6 +64,7 @@ require (
 	github.com/nxadm/tail v1.4.8 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.0.2-0.20210730191737-8e42a01fb1b7 // indirect
+	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/prometheus/client_golang v1.11.0 // indirect
 	github.com/prometheus/client_model v0.2.0 // indirect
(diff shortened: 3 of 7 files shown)


controllers/imagerepository_controller.go
```diff
@@ -240,6 +247,40 @@ func getAwsECRLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error
 	return authConfig, nil
 }
 
+func geGCRLoginAuth() (authn.AuthConfig, error) {
+	var authConfig authn.AuthConfig
+	gcpDefaultTokenURL := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+
+	request, err := http.NewRequest("GET", gcpDefaultTokenURL, nil)
```
relu, line 254: Nice, so we don't even need a client lib for this :ok_hand:

relu:
This is a note for later but I think it would be better to get these implementations out of the controller, maybe moved to `internal`.

controllers/imagerepository_controller.go
```diff
@@ -291,6 +346,46 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1
 		} else {
 			logr.FromContext(ctx).Info("No image credentials secret referenced, and ECR authentication is not enabled. To enable, set the controller flag --aws-autologin-for-ecr")
 		}
+	} else if strings.Contains(imageRepo.Spec.Image, "gcr.io") {
```
hiddeco, line 349: If someone decides to register `abcdgcr.io` and host images on it, this is going to magically fail for them.

internal/azure/exchanger.go
```diff
@@ -0,0 +1,81 @@
...
+	req, err := http.NewRequest("POST", exchangeUrl, strings.NewReader(parameters.Encode()))
+	if err != nil {
+		return "", fmt.Errorf("error constructing exchange request: %w", err)
+	}
+
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Add("Content-Length", strconv.Itoa(len(parameters.Encode())))
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
relu, line 54: There's a helper method that can be used for form POSTing in `http`. Here's the short variant of all of this :smile: 
```suggestion
	resp, err := http.PostForm(exchangeUrl, parameters)
```

internal/azure/exchanger.go
```diff
@@ -0,0 +1,81 @@
...
+
+type tokenResponse struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	Resource     string `json:"resource"`
+	TokenType    string `json:"token_type"`
+}
+
+type Exchanger struct {
+	scheme   string
```
relu, line 21: Doesn't seem like something that would ever be changed or overridden, should be safe to hardcode this as `https` in the resulting URL.

controllers/imagerepository_controller.go
```diff
@@ -240,6 +260,45 @@ func getAwsECRLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error
...
+// getGCRLoginAuth obtains authentication for the image by
+// getting a token from the metadata API on GCP. This assumes that
+// the pod has right to pull the image which would be the case if it
+// is hosted on GCP. It works with both service account and workload identity
+// enabled clusters.
+func getGCRLoginAuth() (authn.AuthConfig, error) {
+	var authConfig authn.AuthConfig
+	gcpDefaultTokenURL := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+
+	request, err := http.NewRequest("GET", gcpDefaultTokenURL, nil)
```
relu, line 272: Maybe use the const here
```suggestion
	request, err := http.NewRequest(http.MethodGet, gcpDefaultTokenURL, nil)
```

controllers/imagerepository_controller.go
```diff
@@ -585,3 +684,45 @@ func getURLHost(urlStr string) (string, error) {
...
+	}
+
+	ex := azure.NewExchanger(ref.Context().RegistryStr())
+	accessToken, err := ex.ExchangeACRAccessToken(string(armToken.Token))
+	if err != nil {
+		return authConfig, fmt.Errorf("error exchanging token: %w", err)
+	}
+
+	return authn.AuthConfig{
+		Username: "00000000-0000-0000-0000-000000000000",
```
relu, line 711: Could you please add a comment regarding this value? Looks like a placeholder, would be great to understand why it's set to this.

relu (changes requested):
Looking good, @SomtochiAma :+1: 
Added some comments regarding a few minor improvements.

internal/azure/exchanger.go
```diff
@@ -0,0 +1,67 @@
...
+	parameters.Add("grant_type", "access_token")
+	parameters.Add("service", parsedURL.Hostname())
+	parameters.Add("access_token", armToken)
+
+	resp, err := http.PostForm(exchangeUrl, parameters)
+	if err != nil {
+		return "", fmt.Errorf("failed to send token exchnage request: %w", err)
+	}
+
+	if resp.StatusCode != 200 {
```
relu, line 45:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

internal/azure/exchanger.go
```diff
@@ -0,0 +1,67 @@
...
+		responseBytes, err := ioutil.ReadAll(resp.Body)
```
relu, line 46: Might be a good idea to not use `ioutil.ReadAll` here, we don't know how large the response body can be and this can cause memory issues. See my comment below.

Do we know what type of responses we get? If it's JSON maybe we can extract only the relevant bits for the error message. If not, maybe try to only read at a maximum just a few bytes worth of data.

internal/azure/exchanger.go
```diff
@@ -0,0 +1,67 @@
...
+			resp.StatusCode, string(responseBytes))
+	}
+
+	responseBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	var tokenResp tokenResponse
+	err = json.Unmarshal(responseBytes, &tokenResp)
```
relu, line 61: `ioutil.ReadAll` loads everything into memory and it should be avoided whenever possible. Using the json Decoder would be more efficient.
```suggestion
        var tokenResp tokenResponse
        decoder := json.NewDecoder(resp.Body)
        if err := decoder.Decode(&tokenResp); err != nil {
        	return "", err
        }
```

internal/azure/exchanger.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func NewExchanger(acrEndpoint string) *Exchanger {
+	return &Exchanger{
+		acrFQDN: acrEndpoint,
+	}
+}
+
+func (e *Exchanger) ExchangeACRAccessToken(armToken string) (string, error) {
+	exchangeUrl := fmt.Sprintf("%s://%s/oauth2/exchange", "https", e.acrFQDN)
```
relu, line 29:
```suggestion
	exchangeUrl := fmt.Sprintf("https://%s/oauth2/exchange", e.acrFQDN)
```

relu (changes requested):
Left just a few more comments.

kingdonb:
I've done some testing against the image `somma/image-reflector-controller:test-autologin-c34c7cf34b1` on both Azure and GKE, and the feature was working all around for me.

```
$ flux get image policy -A
NAMESPACE       NAME    READY   MESSAGE                                                                                         LATEST IMAGE                                         
podinfo-staging podinfo True    Latest image tag for 'acrkingdontest.azurecr.io/stefanprodan/podinfo' resolved to: 5.0.0        acrkingdontest.azurecr.io/stefanprodan/podinfo:5.0.0
```

and on GCP:
```
$ flux get image policy -A
NAMESPACE               NAME    READY   MESSAGE                                                                 LATEST IMAGE                    
podinfo-production      podinfo True    Latest image tag for 'gcr.io/dx-kingdon/podinfo' resolved to: 5.0.0     gcr.io/dx-kingdon/podinfo:5.0.0
```

The testing involved was to follow basically the first half of the Image Update guide: create an ImageRepo and ImagePolicy, pointed at the private repository. Remove the flag to confirm that access does not happen without the newly added feature. When the flag is present, there should be no secretRef or other access management required, it should "just work."

I created the Azure cluster in an existing Resource Group through the portal with a "System-Managed Identity" and associated the cluster with a Container Registry on the same RG. I used Azure CNI and Calico policy enforcement, which probably doesn't matter, (but just to be complete) with Kubernetes 1.21 under the hood. The image repo created was definitely private. I used the Quickstart guide from Azure docs to "build and push" some example Podinfo images `FROM stefanprodan/podinfo:5.0.0`.

https://docs.microsoft.com/en-us/azure/container-registry/container-registry-quickstart-task-cli

The Dockerfiles for example are in here: https://github.com/kingdonb/bootstrap-repo/tree/main/apps/base/podinfo - they are very sparse. I just needed two images so I could confirm they were both discovered successfully when the ImageRepo is reconciled.

I did the same test on GCR basically, creating a GKE cluster with Workload Identity enabled, then building the same images from the tutorial for Google Cloud Build. This testing method made it possible for me to avoid using my local network or any local build nodes 👍 I placed the images in a new repository in the same `project-id` that my cluster was created in. I assume this is using the new "Artifact Registry" protocol but I haven't looked deeply into it yet. (Maybe separate validation is still needed for that.) I just used the most obvious docs and searched for anything that I didn't know, I'm only marginally familiar with both cloud vendors.

https://cloud.google.com/build/docs/building/build-containers

The Workload Identity automatically authorized the project repository in the same workspace (project) and as I was able to push images with the gcloud client, that were then discovered without issue by the image-reflector-controller, with a set of manifests in and centered around this one: https://github.com/kingdonb/bootstrap-repo/blob/main/apps/production/podinfo/imagerepo.yaml

After confirming these both worked successfully with the flags:

https://github.com/kingdonb/bootstrap-repo/blob/75fc7f8822f1850d4046700ce4281c39ac5e4bfd/clusters/gke-geekingdon/flux-system/kustomization.yaml#L16
and
https://github.com/kingdonb/bootstrap-repo/blob/75fc7f8822f1850d4046700ce4281c39ac5e4bfd/clusters/aks-kuberkingdon/flux-system/kustomization.yaml#L16

I'm happy to report the feature works without issue on both of these! Removing the flag made for an authentication error that looked reasonable, (I didn't capture it.) I'll be repeating the testing later today, in case there are any questions or issues in the critical path that other reviewers wanted to make sure are properly addressed. Please let me know.

internal/azure/exchanger.go
```diff
@@ -0,0 +1,67 @@
+package azure
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
```
stefanprodan, line 6: The `io/ioutil` package is deprecated, please use `io`.

stefanprodan (changes requested):

relu (approved):
Great, thank you @SomtochiAma! :100:

relu:
Could you please also resolve the conflicts?

kingdonb:
I have some surprising results after testing further on GKE,

I created a new cluster and enabled Workload Identity, but the reflector controller can no longer reach the auth, or something else has gone wrong:

```
{"level":"error","ts":"2021-11-23T20:02:18.456Z","logger":"controller-runtime.manager.controller.imagepolicy","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImagePolicy","name":"podinfo","namespace":"podinfo-production","error":"version list argument cannot be empty"}
{"level":"info","ts":"2021-11-23T20:02:21.686Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Logging in to GCP GCR for gcr.io/dx-kingdon/podinfo","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"podinfo","namespace":"podinfo-production"}
{"level":"error","ts":"2021-11-23T20:02:21.878Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"podinfo","namespace":"podinfo-production","error":"GET https://gcr.io/v2/dx-kingdon/podinfo/tags/list?n=1000: DENIED: Failed to read tags for host 'gcr.io', repository '/v2/dx-kingdon/podinfo/tags/list?n=1000'"}
{"level":"info","ts":"2021-11-23T20:03:59.667Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Logging in to GCP GCR for gcr.io/dx-kingdon/podinfo","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"podinfo","namespace":"podinfo-production"}
{"level":"error","ts":"2021-11-23T20:03:59.693Z","logger":"controller-runtime.manager.controller.imagepolicy","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImagePolicy","name":"podinfo","namespace":"podinfo-production","error":"version list argument cannot be empty"}
{"level":"error","ts":"2021-11-23T20:03:59.875Z","logger":"controller-runtime.manager.controller.imagerepository","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImageRepository","name":"podinfo","namespace":"podinfo-production","error":"GET https://gcr.io/v2/dx-kingdon/podinfo/tags/list?n=1000: DENIED: Failed to read tags for host 'gcr.io', repository '/v2/dx-kingdon/podinfo/tags/list?n=1000'"}
```

This while the images can be deployed in the cluster, so apparently the project association is correct, something else has gone wrong.

I suspect this is probably a configuration error, although I'm not sure what can have gone wrong, I have the flag enabled and my image is at:

```
images: # []
- name: ghcr.io/fluxcd/image-reflector-controller
  newName: docker.io/somma/image-reflector-controller
  newTag: test-autologin-c34c7cf34b1
```

but the registry cannot be scanned for ImageRepository resources. Is there somewhere else to check for debugging errors?

kingdonb:
@SomtochiAma sent me some docs about Workload Identity that I hadn't seen before. It looks like my previous test was not using workload identity, but the default service account instead. I will go back and repeat it after I work out my Workload Identity issues to confirm. Anyway just to add to this report, my configuration issue is definitely a configuration issue.

SomtochiAma:
I am adding some notes here for testing this on GCP. I will write more detailed documentation next and make a commit.

1. Enable Workload Identiy on the cluster (Docs: [https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_on_cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_on_cluster). If you have an existing cluster, don't forget to migrate applications to workload identity. See [https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#migrate_applications_to](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#migrate_applications_to)
2. Create a GCP service account that has access to GCR
3. Bind the GCP service account to the image-reflector-controller service account ([Docs](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to))
4. [Customize](https://fluxcd.io/docs/get-started/#customize-podinfo-deployment) the flux manifest so that it uses the image from this pull request and also annotates the image-reflector-controller service account with the ` iam.gke.io/gcp-service-account` label

Example of `gotk-patches.yaml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: image-reflector-controller
  namespace: flux-system
spec:
  template:
    spec:
      containers:
      - name: manager
        image: somma/image-reflector-controller:test-autologin-d5255a590
        args:
          - --events-addr=http://notification-controller.flux-system.svc.cluster.local/
          - --watch-all-namespaces
          - --log-level=info
          - --log-encoding=json
          - --enable-leader-election
          - --gcp-autologin-for-gcr
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: image-reflector-controller
  namespace: flux-system
  annotations:
    iam.gke.io/gcp-service-account: "flagger-stackdriver@dx-somtochi.iam.gserviceaccount.com"
```

stefanprodan:
The bootstrap docs along with #193 should be placed in https://fluxcd.io/docs/guides/image-update/#imagerepository-cloud-providers-authentication

squaremo (approved):
Solid work @SomtochiAma, thank you. I've made a few suggestions and raised some questions -- nothing drastic, so I've approved this PR.

(The first commit seems to have a duplicate sign-off -- something to amend if you end up doing further work here.)

.github/workflows/scan.yml
```diff
@@ -27,6 +27,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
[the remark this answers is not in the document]
SomtochiAma: The azure package has build constraints for the go version. The CodeQL version was lower  and it caused the analysis to fail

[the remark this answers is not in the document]
squaremo: Got it -- it's worth leaving a comment, or mentioning in the commit message, why this is so.

.github/workflows/scan.yml
```diff
@@ -27,6 +27,9 @@ jobs:
...
+        with:
+          go-version: '1.16'
```
stefanprodan, line 32:
```suggestion
          go-version: '1.17.x'
```
To match https://github.com/fluxcd/image-reflector-controller/pull/190

go.mod
```diff
@@ -22,6 +24,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.11.0
 )
 
+require github.com/Azure/azure-sdk-for-go/sdk/internal v0.8.1 // indirect
```
stefanprodan, line 27: Please copy this indirect to the section below

internal/azure/exchanger.go
```diff
@@ -0,0 +1,66 @@
+package azure
+
```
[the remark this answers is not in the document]
stefanprodan: We need to copy the copyright headers from the original project and link to it. We also need to add our own copyright and explain what we changed. See here an example: https://github.com/fluxcd/pkg/blob/main/ssh/knownhosts/knownhosts.go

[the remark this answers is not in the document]
SomtochiAma: The original project has no copyright headers so I will just link it.

kingdonb:
It's noted elsewhere, but following up that I have tested GKE Workload Identity to connect with gcr.io registries successfully

However I was not able to get artifact registry working through workload identity, even after separately granting the role for `role/artifactregistry.reader` I still get this error:

```
podinfo-artifact-registry	False	GET https://us-central1-docker.pkg.dev/v2/token?scope=repository%3Adx-kingdon%2Fgeekingdon-images%2Fpodinfo%3Apull&service=us-central1-docker.pkg.dev: DENIED: Permission "artifactregistry.repositories.downloadArtifacts" denied on resource "projects/dx-kingdon/locations/us-central1/repositories/geekingdon-images" (or it may not exist)	                         	False
```

I guess the culprit is here: https://github.com/fluxcd/image-reflector-controller/pull/194/files#diff-922fcbfbe2565443329918a2b0d49b2b03588007ec6464a20d3feb824f3f80a2R735-R737

it needs to match `pkg.dev` as well (and possibly others), else artifact registry won't be considered as the same as container registry. I'm giving it a shot and testing it on my own, I think I have everything I need to compile a version of image-reflector-controller 👍

kingdonb:
The other thing I might want to add to the docs:

```
gcloud projects add-iam-policy-binding <PROJECT> --member "serviceAccount:<GCP-SA-NAME>@<PROJECT>.iam.gserviceaccount.com" --role roles/artifactregistry.reader
gcloud projects add-iam-policy-binding <PROJECT> --member "serviceAccount:<GCP-SA-NAME>@<PROJECT>.iam.gserviceaccount.com" --role roles/containerregistry.ServiceAgent
```

It was not very difficult to discover these roles, but we can help the users by spelling it out.

I'm not sure I have selected the minimal roles either. There is no `containerregistry.reader` and after applying both (and possibly some others), I wind up with this set of bindings which seems like it might likely be in excess of what I needed:

```
Updated IAM policy for project [dx-kingdon].
bindings:
- members:
  - serviceAccount:kingdon-test-gcr-sa@dx-kingdon.iam.gserviceaccount.com
  role: roles/artifactregistry.reader
- members:
  - serviceAccount:service-1041773995540@gcp-sa-artifactregistry.iam.gserviceaccount.com
  role: roles/artifactregistry.serviceAgent
- members:
  - serviceAccount:1041773995540@cloudbuild.gserviceaccount.com
  role: roles/cloudbuild.builds.builder
- members:
  - serviceAccount:service-1041773995540@gcp-sa-cloudbuild.iam.gserviceaccount.com
  role: roles/cloudbuild.serviceAgent
- members:
  - serviceAccount:service-1041773995540@compute-system.iam.gserviceaccount.com
  role: roles/compute.serviceAgent
- members:
  - serviceAccount:service-1041773995540@container-engine-robot.iam.gserviceaccount.com
  role: roles/container.serviceAgent
- members:
  - serviceAccount:kingdon-test-gcr-sa@dx-kingdon.iam.gserviceaccount.com
  - serviceAccount:service-1041773995540@containerregistry.iam.gserviceaccount.com
  role: roles/containerregistry.ServiceAgent
- members:
  - serviceAccount:1041773995540-compute@developer.gserviceaccount.com
  - serviceAccount:1041773995540@cloudservices.gserviceaccount.com
  role: roles/editor
- members:
  - user:kingdon@weave.works
  role: roles/owner
- members:
  - serviceAccount:service-1041773995540@gcp-sa-pubsub.iam.gserviceaccount.com
  role: roles/pubsub.serviceAgent
etag: BwXVjQ36L8E=
version: 1
```

Someone who knows GCP well enough to say how to select the minimal roles should review that information for correctness.

stefanprodan:
@SomtochiAma can you please rebase and run `make fmt`.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🏅

controllers/imagerepository_controller.go
```diff
@@ -246,6 +266,45 @@ func getAwsECRLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error
...
+	}
+
+	request.Header.Add("Metadata-Flavor", "Google")
+
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		return authConfig, err
+	}
+
```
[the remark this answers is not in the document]
squaremo: You should really do this in all cases other than getting an error back from `Do`. As it stands, the body will only be read to the end and closed if the status was OK and a JSON value could be decoded; but, it's entirely possible, e.g., that a 40x status would have body content.
It's a pain I know, because you are forced to either repeat code or defer (and throw away errors) -- since no (other) I/O happens after the request, I think using `defer` is fine.

squaremo (changes requested):
Thanks for following through on my comments -- one adjustment, then I think this is good to go.

stefanprodan:
@SomtochiAma can you please open a new PR to address @squaremo comment, sorry I missed it.

PR #196, opened 2021-11-11
Watched cross-ns image repos trigger reconcile

### LINKED ISSUE #195: ImagePolicy resources not reconciled by watched cross-namespace ImageRepositories the refer
> I think you might have namespaces of the ImageRepository and ImagePolicy backwards - assuming your GitRepository and ImageUpdateAutomation objects are in the flux-system namespace, then the ImagePolicy should also be in the flux-system namespace and the ImageRepository should be in the "app" namespace.
 
> But even after making that change I am having an issue with the ImagePolicy not reconciling the image tags that the ImageRepository is pulling. After playing around with it for a while, I noticed that restarting the image reflector controller will cause the cross namespace ImagePolicy to be reconciled exactly once, but any new tags pushed to the image repository are completely ignored. Looking through the image reflector controller logs (level=debug) I confirmed that the controller will only reconcile ImagePolicy objects using cross namespace references shortly after the controller starts up and then completely ignores them unlike the ImagePolicy objects that are NOT using the cross namespace references. I'm thinking either the documentation is missing something extremely crucial in order to get this to work or there is a bug. Also I apologize in advance for not sharing the logs as they are work related.

_Originally posted by @e-dang in https://github.com/fluxcd/flux2/discussions/1958#discussioncomment-1622099_

relu:
Cross-namespace ImageRepository resources should trigger reconciles for
ImagePolicies that refer to them. Previously, this was only done for
resources in the same namespace.

Fixes #195

### DIFF
--- a/controllers/imagepolicy_controller.go
+++ b/controllers/imagepolicy_controller.go
@@ -48,7 +48,7 @@ import (
 // this is used as the key for the index of policy->repository; the
 // string is arbitrary and acts as a reminder where the value comes
 // from.
-const imageRepoKey = ".spec.imageRepository.name"
+const imageRepoKey = ".spec.imageRepository"
 
 // ImagePolicyReconciler reconciles a ImagePolicy object
 type ImagePolicyReconciler struct {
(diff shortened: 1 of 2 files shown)


controllers/policy_test.go
```diff
@@ -687,6 +688,19 @@ var _ = Describe("ImagePolicy controller", func() {
 				}, timeout, interval).Should(BeTrue())
 				Expect(pol.Status.LatestImage).To(Equal(imgRepo + ":1.0.1"))
 
+				// Updating the image should reconcile the cross-namespace policy
```
squaremo, line 691: :+1:

controllers/imagepolicy_controller.go
```diff
@@ -259,8 +259,7 @@ func (r *ImagePolicyReconciler) SetupWithManager(mgr ctrl.Manager, opts ImagePol
 func (r *ImagePolicyReconciler) imagePoliciesForRepository(obj client.Object) []reconcile.Request {
 	ctx := context.Background()
 	var policies imagev1.ImagePolicyList
-	if err := r.List(ctx, &policies, client.InNamespace(obj.GetNamespace()),
-		client.MatchingFields{imageRepoKey: obj.GetName()}); err != nil {
+	if err := r.List(ctx, &policies, client.MatchingFields{imageRepoKey: obj.GetName()}); err != nil {
```
squaremo, line 262: This can result in collisions unless it includes the namespace (and the indexing should follow suit).

relu: Yes, @stefanprodan pointed this out and I'm working on it.

stefanprodan (approved):
LGTM

Thanks @relu

PR #218, opened 2022-01-12
Fix makefile envtest and controller-gen usage

aryan9600:
Refactor logic to install helper tools into one function in the
Makefile. Add support for envtest to help install tools like kubectl,
etcd which helps users run tests more conveniently.

Signed-off-by: Sanskar Jaiswal <sanskar.jaiswal@weave.works>

Ref: https://github.com/fluxcd/flux2/issues/2273

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -38,12 +38,8 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1.17.x
-    - name: Set up kubebuilder
-      uses: fluxcd/pkg/actions/kubebuilder@main
     - name: Run tests
       run: make test
-      env:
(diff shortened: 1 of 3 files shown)


stefanprodan:
@aryan9600 it should be safe now to remove:
- https://github.com/fluxcd/image-reflector-controller/blob/main/.github/workflows/build.yaml#L41
- https://github.com/fluxcd/image-reflector-controller/blob/main/.github/workflows/build.yaml#L46

squaremo:
@aryan9600 Please would you explain the change in the commit message: what needed to be fixed? How does the commit address that? (Some commits are self-explanatory, but I don't think this one is)

stefanprodan (approved):
LGTM

Thanks @aryan9600

PR #219, opened 2022-01-14
Clarify ECR Auto-login note, fix CanonicalImageName typos

kingdonb:
Leans on:

* #194,
(but I think they can merge in either order)

Slimmed-down replacement version of #193 that takes into account most other things that were to be fixed here, are fixed by fluxcd/website#702

There was a usability issue reported in the docs that I meant to address here, and it's not obvious from the structure of the change, but as the `ImageRepositories` controller API docs are not obviously associated with the `image-reflector-controller` where they are most accessible (on the website, [here](https://fluxcd.io/docs/components/image/imagerepositories/#ecr-and-eks) in the API docs)

There is no breadcrumb which says `image-reflector-controller` visible anywhere on the page, so it's necessary to spell out which controller needs to get the flag.

In fluxcd/website#702 a detailed guide is added that covers ECR, as well as the other cloud providers, and so this PR can omit the example of how to set up the patch in deference to a link to the guide instead. 👍

### DIFF
--- a/docs/spec/v1beta1/imagerepositories.md
+++ b/docs/spec/v1beta1/imagerepositories.md
@@ -86,6 +86,14 @@ the flag is `--gcp-autologin-for-gcr`.
 For [<abbr title="Azure Kubernetes Service">AKS</abbr>][AKS] and [<abbr title="Azure Container Registry">ACR</abbr>][ACR],
 the flag is  `--azure-autologin-for-acr`.
 
+These flags can be added by including a patch in the `kustomization.yaml` overlay file in your `flux-system`,
+as described in [cloud providers authentication guide][]. If there is no need for a security boundary on your
+cluster around container registries and you are not using Flux with so-called "soft multi-tenancy", then
+you will likely prefer to use the Auto-Login feature for the convenience and improved ease of use.
+
+Alternatively, the advice to use a cron job to refresh a secret token under [Other platforms][other platforms]
+below will also work with ECR, GCR, and ACR environments that require security boundaries and soft multi-tenancy.
+
 #### Other platforms
 
 If you are running on another platform that links service permissions to service accounts, you will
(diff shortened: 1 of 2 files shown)


docs/spec/v1beta1/imagerepositories.md
```diff
@@ -77,9 +77,17 @@ For a publicly accessible image repository, you will not need to provide a `secr
...
+
+The `image-reflector-controller` must be run with the flag `--aws-autologin-for-ecr` set for this to work.
+
+This flag can be added by including a patch in the `kustomization.yaml` overlay file in your `flux-system`,
+as described in [Using Native AWS ECR Auto-Login][]. If there is no need for a security boundary on your
+cluster around container registries and you are not using Flux with so-called "soft multi-tenancy", then
+you will likely prefer to use the Auto-Login feature for the convenience and improved ease of use.
+
+Alternatively, the advice to use a cron job to refresh a secret token under [Other platforms][other platforms]
+below will also work with ECR for any environment that requires security boundaries and soft multi-tenancy.
```
kingdonb, line 90: This might still be too much bootstrap-centric. If we prefer, I can shorten this to only say what was on line 82.

I'd still like to drop the link to `[Using Native AWS ECR Auto-Login][]` somewhere but otherwise think maybe could do without this entire paragraph line 84-87, it's important information but does it belong in the API doc?

I would merge it anyway if anyone says I'm being too picky, but still welcome feedback about this and anything else.

kingdonb: Then again, for balance maybe this whole section goes away. We won't have a similar section for GCP or AKS here, I think (?)

kingdonb:
It will be hard to make more progress on this PR before fluxcd/website#702 is merged (and in turn, image-reflector-controller#194), so it can be reviewed in context.

docs/spec/v1beta1/imagerepositories.md
relu: Since most of this information is already present under the ["Using Native AWS ECR Auto-Login" section](https://fluxcd.io/docs/guides/image-update/#using-native-aws-ecr-auto-login) in the "Automate image updates to Git" guide wouldn't it make sense to just extend that section and maybe just refer to it from here?

relu (changes requested):

kingdonb:
Rebased to get this in shape to merge 👍

docs/spec/v1beta1/imagerepositories.md
kingdonb: I have adopted this suggestion in the latest version 👍

PR #222, opened 2022-01-18
Fix scheme validation check when using host:port

### LINKED ISSUE #220: Container Registry on 5005 port
Hello,

I have deployed **image-reflector-controller** with version: **ghcr.io/fluxcd/image-reflector-controller:v0.15.0**
When I want to create ImageRepository with port 5005, eg. git.mydomain.com:5005/foo/bar I got message:

`
✗ .spec.image value should not start with URL scheme; remove 'git.mydomain.com://'
`

How to fix this? 

Ps. Version v0.14.0 works fine.

relu:
The validation check for the presence of a scheme did not take into
account situations when the registry hostname would be accompanied by a
port number. In this situation, the hostname would be erroneously parsed
as a scheme and validation would fail.

Fixes #220

### DIFF
--- a/controllers/scan_test.go
+++ b/controllers/scan_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/http/httptest"
+	"strings"
 
 	"github.com/fluxcd/pkg/apis/meta"
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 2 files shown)


controllers/scan_test.go
```diff
@@ -311,5 +312,35 @@ var _ = Describe("ImageRepository controller", func() {
 			}, timeout, interval).Should(BeTrue())
 			Expect(ready.Message).To(ContainSubstring("should not start with URL scheme"))
 		})
+		It("does not fail if using a hostname with a port number", func() {
+			imgRepo := loadImages(registryServer, "test-fetch", []string{"1.0.0"})
+			imgRepo = strings.ReplaceAll(imgRepo, "127.0.0.1", "localhost")
```
relu, line 317: I didn't realize the behavior would change when dealing with hostnames vs an IP address. Here's a test to cover that situation.

stefanprodan: Is there a port in number in this test?

squaremo: I don't think it is -- https://play.golang.com/p/0Vw2HyPHb4P. Now I don't know how it worked before :-S

relu: @stefanprodan there is although not explicitly, the `registryServer` listens on a random port which ends up being included in the URI (e.g. `127.0.0.1:33571/test-fetch`).

@squaremo it does change, try with `localhost:5000/bar` which will succeed parsing but will fail to identify the (missing) scheme. The IP address format will just fail to parse. It worked before because we didn't look at the error but rather just checked if the parsing succeeded and if there was a `Scheme` attribute set (which was not correctly set in the situation when a port would be part of the URI when using a proper hostname).

squaremo: I see, `localhost:5000/bar` parses as a URL with scheme of `localhost` and a path, so falls foul of that condition; but `127.0.0.1:5000` doesn't parse as a URL, and skips ahead.

squaremo (approved):
From eyeballing this, I think it catches the special case (a user supplies an image repo string starting with "http://" or "https://"), and won't accidentally catch e.g., `localhost:5000/bar` as reported in #220.

It might be worth factoring the image repo validation into its own func and testing / fuzzing that directly. But, up to you :-) Thanks @relu 🍏

relu:
Thaks, @squaremo! I'm actually looking into some improvements on this side and will take your suggestion into consideration.

stefanprodan:
I'm for including this in the next release. @relu could you create an issue for fuzzing so we can merge this PR as it is? Or you prefer to add fuzzing this week?

relu:
@stefanprodan I'd like to add fuzzing and might be able to get to it tomorrow.

stefanprodan (approved):
LGTM

Thanks @relu, please rebase.

PR #224, opened 2022-01-20
GCR fix: Discard and close body regardless of status code

SomtochiAma:
A little follow-up on the #194 to ensure the response body is closed regardless of status code


squaremo:
(I think the CI is failing because there's a stray space that gets gofmt'd out, meaning the "is the working tree dirty?" check fails)

squaremo (approved):
Yup looks right, thanks Somtochi.

PR #226, opened 2022-01-24
Move squaremo to retired maintainers

squaremo:
This moves me into the "Retired maintainers" category. The format is intended to work with the script https://github.com/fluxcd/community/blob/main/project/get_project_data.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -2,11 +2,12 @@ The maintainers are generally available in Slack at
 https://cloud-native.slack.com in #flux (https://cloud-native.slack.com/messages/CLAJ40HV3)
 (obtain an invitation at https://slack.cncf.io/).
 
-In additional to those listed below, this project shares maintainers
-from the main Flux v2 git repository, as listed in
+This project shares maintainers from the main Flux v2 git repository,
+as listed in
 
     https://github.com/fluxcd/flux2/blob/main/MAINTAINERS
 
-In alphabetical order:
+Retired maintainers:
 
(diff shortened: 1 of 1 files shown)


MAINTAINERS
```diff
@@ -2,11 +2,16 @@ The maintainers are generally available in Slack at
...
+In additional to those listed as current maintainers below, this project shares maintainers
 from the main Flux v2 git repository, as listed in
 
     https://github.com/fluxcd/flux2/blob/main/MAINTAINERS
 
-In alphabetical order:
+Current project maintainers, in alphabetical order:
 
-Michael Bridgen, Weaveworks <michael@weave.works> (github: @squaremo, slack: Michael Bridgen)
+None.
```
kingdonb, line 12: Echo the sentiment from https://github.com/fluxcd/image-automation-controller/pull/300#pullrequestreview-861499775

We ought not have zero maintainers, if we can avoid it. I'm willing to volunteer (although I'm not sure my contributions here qualify me to be elected as a maintainer) but as a first order of business, I'll be calling attention to the fact that we ought not have only one maintainer, if we can avoid it! :) Ideally a few people can volunteer to make this a smooth transition.

Thanks a bunch @squaremo !

stefanprodan: This project hasn't been abandoned by the Flux team. It is maintained by all these people: https://github.com/fluxcd/flux2/blob/main/MAINTAINERS

relu (approved):

PR #227, opened 2022-01-25
Publish SBOM and sign release artifacts

stefanprodan:
This PR implements:
- https://github.com/fluxcd/flux2/issues/2302
- https://github.com/fluxcd/flux2/issues/2303

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,11 @@ on:
         default: 'rc'
         required: true
 
+permissions:
+  contents: write # needed to write releases
+  id-token: write # needed for keyless signing
+  packages: write # needed for ghcr access
+
 env:
   CONTROLLER: ${{ github.event.repository.name }}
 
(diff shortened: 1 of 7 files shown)


darkowlzz (approved):
LGTM!

.github/workflows/release.yml
```diff
@@ -49,43 +50,51 @@ jobs:
...
           mkdir -p config/release
           kustomize build ./config/crd > ./config/release/${{ env.CONTROLLER }}.crds.yaml
           kustomize build ./config/manager > ./config/release/${{ env.CONTROLLER }}.deployment.yaml
-      - name: Create release
+          echo '[CHANGELOG](https://github.com/fluxcd/${{ env.CONTROLLER }}/blob/main/CHANGELOG.md)' > ./config/release/notes.md
+      - uses: anchore/sbom-action/download-syft@v0
+      - name: Create release and SBOM
         if: startsWith(github.ref, 'refs/tags/v')
-        uses: ncipollo/release-action@v1
+        uses: goreleaser/goreleaser-action@v2
```
relu, line 95: I couldn't figure out if this action can be configured to mark the release as a prerelease as was the case with `ncipollo/release-action@v1`. I assume we still want prereleases until Flux goes GA, right?

stefanprodan: Yes, the prerelease flag is set in the .goreleaser.yaml file.

relu (approved):

PR #228, opened 2022-01-27
Allow disabling cross-namespace references to image repositories

squaremo:
Introduce the flag `--no-cross-namespace-refs` (defaults to false) for allowing cluster admins to disable cross-namespace references to `ImageRepositories`.

When the controller is run with `--no-cross-namespace-refs=true` and a `ImagePolicy.spec.imageRepositoryRef.namespace` refers to an `ImageRepository` in a different namespace than the `ImagePolicy` object, the reconciliation will fail with the `AccessDenied` reason.

On access denied errors the controller logs and sends an event e.g.:

```
can't access 'ImageRepository/my-other-namespace/my-source', cross-namespace references have been blocked
```

And the `ImagePolicy` status `Ready` condition it set to:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-01-26T07:26:48Z"
    message: "can't access 'ImageRepository/my-other-namespace/my-source', cross-namespace references have been blocked"
    reason: AccessDenied
    status: "False"
    type: Ready
```

Part of: https://github.com/fluxcd/flux2/issues/2337


stefanprodan (approved):
LGTM

Thanks @squaremo 🥇

PR #229, opened 2022-01-28
Allow custom port to image registry

cseccia:
The parse method works bad with port without schema.
If applied this commit will permit to use format [path]:[port]/[owner]/[registry] for image spec


cseccia:
Thanks and sorry for the duplication

stefanprodan:
Duplicate of in #222

PR #231, opened 2022-02-14
fix: Invalid Azure Scope

s-bauer:
The scope is missing a `/.default` at the end and thereby prevented me from acquiring a valid access token.

Azure requires a full scope and not only the URL to be able to acquire a token. As the flow uses application permissions and not delegate permissions, the scope needs to end in "/.default" as described in https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#request-the-permissions-in-the-app-registration-portal

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -687,7 +687,7 @@ func getAzureLoginAuth(ctx context.Context, ref name.Reference) (authn.AuthConfi
 		return authConfig, err
 	}
 	armToken, err := cred.GetToken(ctx, policy.TokenRequestOptions{
-		Scopes: []string{string(arm.AzurePublicCloud)},
+		Scopes: []string{string(arm.AzurePublicCloud) + ".default"},
 	})
 	if err != nil {
 		return authConfig, err


relu:
Thank you, @s-bauer! Would you mind amending your commit to resolve the [DCO](https://github.com/fluxcd/image-reflector-controller/pull/231/checks?check_run_id=5180096365) check? Maybe also add some more information about this fix, like a link to some relevant documentation.

s-bauer:
@relu Added more information and my sign-off.

SomtochiAma:
Hey @s-bauer,

The scope works okay. What is needed is to install [AAD Pod Identity](https://azure.github.io/aad-pod-identity/docs/getting-started/installation/), create Azure identity and azure identity binding, then add a `aadpodidbinding:<name-of-az-identity>` label to the `image-reflector-controller` deployment podspec. I am in the process of writing and updating the docs for this.
My apologies that this wasn;t clear from the start🙏🏾

s-bauer:
@SomtochiAma Yea I know this is supposed to be used with the pod identity, but as it has several limitations and there is a pod identity v2 coming, I want to avoid using it at the moment. I've applied a Kustomize patch to the image-reflector-controller to add the `AZURE_CLIENT_ID` `AZURE_TENANT_ID` and `AZURE_CLIENT_SECRET` environment variables. The `DefaultAzureCredentials` use the `EnvironmentCredentials` as the first priority and therefore picks up the credentials correctly.

The only thing that didn't work is the scope as the url ("https://management.azure.com/")  is simply not a valid scope for AAD. I'm not sure why pod identity works with that scope. 

If you have an environment with pod-identity, can you please check if it is also working correctly with the "/.default" suffix?

SomtochiAma:
I can verify that this scope works with aad pod identiy

SomtochiAma (approved):

s-bauer:
Awesome! BTW, just seen about "Azure Workload Identity" which is replacing pod identity and now available in preview: https://azure.github.io/azure-workload-identity/docs/introduction.html

s-bauer:
I've done some tests with the new Azure Workload Identity package and unfortunately it's not yet working with the image-reflector-controller. This is due to the reason that the azure identity library for go does not yet support authentication using federated identities yet (Tracked in https://github.com/Azure/azure-sdk-for-go/issues/15615). A workaround would be to use msal like the example at https://github.com/Azure/azure-workload-identity/tree/main/examples/msal-go or to simply wait until the azure identity sdk for go is updated.

pjbgf:
@s-bauer I caught up with @hiddeco on this and we agreed on merging this PR to fix the Azure Scope issue, and then add support for Azure Workload Identity as a [separate issue](https://github.com/fluxcd/image-reflector-controller/issues/237). 

Can you rebase this PR please?

hiddeco (approved):

hiddeco:
Thank you @s-bauer :bow:

PR #232, opened 2022-02-15
Refactor logging to be more consistent.

aryan9600:
Adds helper functions to help log in a more consistent manner.

Ref: fluxcd/pkg#172

Signed-off-by: Sanskar Jaiswal <jaiswalsanskar078@gmail.com>


pjbgf (approved):
@aryan9600 LGTM

controllers/imagepolicy_controller.go
```diff
@@ -203,35 +177,20 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			err = fmt.Errorf("Cannot determine latest tag for policy: %w", err)
 		}
-		r.event(ctx, pol, events.EventSeverityError, msg)
-		return ctrl.Result{}, nil
+		res, recErr := recordError(err, meta.ReconciliationFailedReason)
+		if recErr != nil {
+			// log the actual error since we are returning the error related to patching status
+			log.Error(err, "")
+			return res, recErr
+		}
```
hiddeco, line 192: Wondering if it would maybe be better to do:

```suggestion
		res, recErr := recordError(err, meta.ReconciliationFailedReason)
		if recErr != nil {
			log.Error(retErr, "failed to record reconciliation failure")
		}
		return ctrl.Result{}, err
```

aryan9600: The `log.Error(err, "")` is actually meant to record the actual error. `recErr` is the error returned if there was an error while actually recording the error. Maybe this could provide more context: https://github.com/fluxcd/image-reflector-controller/commit/046c29e69dd09af0dba1ba48bd3634c628160621#r66637836

hiddeco (approved):

PR #252, opened 2022-05-02
Adds ServiceAccountName for ImageRepository API

### LINKED ISSUE #251: Add ServiceAccountName for ImageRepository Type
ServiceAccount is a great way to support image pull secrets. It follows standard k8s conventions and would be a good addition to the ImageRepository API. 

This will also be consistent with the proposed OCIRepository API ([RFC #2601](https://github.com/fluxcd/flux2/pull/2601)). 

Here is a [link](https://github.com/fluxcd/flux2/pull/2601#discussion_r849498496) to the discussion regarding this enhancement.

cc: @stefanprodan @squaremo

rashedkvm:
Signed-off-by: Rashed Kamal <krashed@vmware.com>

This PR adds ServiceAccountName to ImageRepository API. Added unit test and ran functional tests. Also updated API docs.

Fixes #251


stefanprodan (approved):
LGTM

Thanks @rashedkvm 🏅 

PS. We need to add the new field to docs but this can come in a different PR.

PR #256, opened 2022-05-10
Add `exclusionList` to enable excluding tags based on regex.

### LINKED ISSUE #167: Allow exclusion of cosign objects from the image repository database
When using [cosign](https://github.com/sigstore/cosign) to sign container images, a standalone object is push to the registry. Our image scanner thinks this is a valid container image tag (GHCR thinks that too) and could potentially crash live system if an image policy matches the cosign object.

For example I've signed [flagger-loadtester](https://github.com/orgs/fluxcd/packages/container/package/flagger-loadtester), GHCR shows the signature object as a container image tag:

Screenshot 2021-08-24 at 16 38 27

As expected Flux adds this tag to its database:

```console
$ flux create image repository flagger-loadtester --image ghcr.io/fluxcd/flagger-loadtester --interval 20m

$ flux get images all
NAME                              	READY	MESSAGE                      	LAST SCAN                	SUSPENDED 
imagerepository/flagger-loadtester	True 	successful scan, found 2 tags	2021-08-24T16:08:15+03:00	False 
```

To avoid deploying cosign signatures as container images onto clusters, I propose we add a field to the `ImageRepository` API to allow people to exclude tags based on regex expression, e.g.:

```yaml
kind: ImageRepository
metadata:
  name: flagger-loadtester
  namespace: flux-system
spec:
  image: ghcr.io/fluxcd/flagger-loadtester
  interval: 20m0s
  exclusionList:
    - ".*sig$"
```

aryan9600:
ExlusionList specifies a list of regex expressions, which is used to
filter out unwanted tags from being stored in the database. If the list
is empty, by default all tags ending with `.sig` are ignored.

Fixes #167 

Signed-off-by: Sanskar Jaiswal <jaiswalsanskar078@gmail.com>


Makefile
```diff
@@ -32,7 +32,7 @@ manager: generate fmt vet
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
-	go run ./main.go
+	go run ./main.go --storage-path=./data
```
stefanprodan, line 35: Add `data` to gitignore please

controllers/imagerepository_controller.go
```diff
@@ -479,15 +482,36 @@ func (r *ImageRepositoryReconciler) scan(ctx context.Context, imageRepo *imagev1
 		)
 		return err
 	}
+	fmt.Printf("tags: %v", tags)
```
stefanprodan, line 485: Please remove all `fmt.Printf`.

stefanprodan (commented):
We need tests to make sure the default is used when no list is provided, and that the default is ignored when a list is provided.

stefanprodan:
You forgot to link to the issue.

docs/spec/v1beta1/imagerepositories.md
```diff
@@ -200,6 +205,13 @@ To grant access to all namespaces, an empty `matchLabels` must be provided:
       - matchLabels: {}
 ```
 
+### Exclude Tags
+
+To exclude certain tags, the `spec.exclusionList` field can be used to specify a list of regex expressions.
+Any tags that match any of the regex expressions, will be excluded from the final tag list.
+If `spec.exclusionList` is empty, by default the regex `"^.*\\.sig$"` is used to exclude all tags ending with
+`.sig`, since these are [Cosign](https://github.com/sigstore/cosign) generated objects and not valid images.
```
stefanprodan, line 213:
```suggestion
`.sig`, since these are [Cosign](https://github.com/sigstore/cosign) generated objects and not container images who can be deployed on a cluster.
```

stefanprodan (approved):
LGTM

Thanks @aryan9600 🏅

Issue #225, opened 2022-01-21
Deleted ImagePolicies and ImageRepositories gotk_reconcile_condition values are incoherent


sa-spag:
👋

I have noticed that deleted ImagePolicies and ImageRepositories conditions metrics are misreported by the controller.

# Steps to reproduce

1. Create a valid ImageRepository and an associated ImagePolicy, e.g. with:
```bash
flux create image repository podinfo \
    --image ghcr.io/stefanprodan/podinfo \
&& \
flux create image policy podinfo \
    --image-ref=podinfo \
    --select-semver="6.x"
```
2. Delete the 2 previously created resources with `kubectl delete imagerepository podinfo && kubectl delete imagepolicy podinfo` (order does not matter much, although it slightly impacts the experienced behavior)

# Expected behavior

- `gotk_reconcile_condition{kind="imagerepository" name="podinfo", type="Ready", status="Deleted"}` and `gotk_reconcile_condition{kind="imagepolicy" name="podinfo", type="Ready", status="Deleted"}` equal 1 for, metrics for other status values equal 0

# Experienced behavior

- `gotk_reconcile_condition{kind="imagerepository" name="podinfo", type="Ready", status="True"}` equals 1, `gotk_reconcile_condition{kind="imagepolicy" name="podinfo", type="Ready", status="False"}` equals 1 too (as the referenced ImageRepository was deleted before the ImagePolicy itself I guess), 
metrics for other status values equal 0

michalschott:
Can confirm, also affected by this.

PR #266, opened 2022-06-02
Add a finalizer to the ImagePolicy and ImageRepository resources

kaden-l-nelson:
Fixes https://github.com/fluxcd/image-reflector-controller/issues/225

This PR adds a finalizer to the ImagePolicy and ImageRepository resources. This is to properly record the Deleted reconciliation status when the object is deleted from the cluster. Without this change, the resource would be deleted before the image reflector controller has a chance to properly report the resource's status in the metrics. As a result, end-users may see falsely reported metrics.

Signed-off-by: Kaden Nelson <kaden_l_nelson@apple.com>

### DIFF
--- a/api/v1beta1/imagepolicy_types.go
+++ b/api/v1beta1/imagepolicy_types.go
@@ -24,6 +24,7 @@ import (
 )
 
 const ImagePolicyKind = "ImagePolicy"
+const ImagePolicyFinalizer = "finalizers.fluxcd.io"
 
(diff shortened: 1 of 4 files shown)


kgamanji:
@stefanprodan I had some community folks query about this PR.  Do you by any chance have any visibility into it or is it on someone else’s radar?

stefanprodan (approved):
LGTM

Thanks @kaden-l-nelson 🏅

PR #273, opened 2022-06-14
Add native registry login e2e test for EKS

darkowlzz:
It introduces a new package `tftestenv`. It is similar to [testenv](https://pkg.go.dev/github.com/fluxcd/pkg/runtime@v0.16.2/testenv) but uses
terraform to create test environment. It provides helpers to run terraform
apply and destroy in the test context.

`tftestenv` is used to create tests similar to the controller tests but with
EKS cluster provisioned using terraform. The test setup is written
similar to the usual controller test suite_test.go, providing
flexibility in the test setup and execution. The structure tries to keep
the infrastructure code separate from the test code such that the
infrastructure provisioning tool can be easily replaced, without the
need to change the actual tests.

The go tests are also written similar to the usual controller tests,
with a kube client from the tftestenv environment global variable.

Terraform is automatically downloaded using the terraform-exec package.
The test container images are generated using the go-containerregistry
package and pushed to the test ECR repository.

`tests/aws/README.md` contains more details about the test setup and workflow.

The terraform configurations for EKS is based on https://github.com/terraform-aws-modules/terraform-aws-eks/tree/6bb73bbdd469797cb067e9aa22ce2f31a7fb22be/examples/self_managed_node_group .

It takes about 25-30 minutes for the whole provisioning, testing and cleanup to finish.


darkowlzz:
Closing for #275 .

PR #275, opened 2022-06-21
Add native registry login tests for EKS, AKS and GKE

darkowlzz:
It introduces a new package `tftestenv`. It is similar to [testenv](https://pkg.go.dev/github.com/fluxcd/pkg/runtime@v0.16.2/testenv) but uses
terraform to create test environment. It provides helpers to run terraform
apply and destroy in the test context.

`tftestenv` is used to create tests similar to the controller tests but with
cloud provider managed cluster provisioned using terraform. The test
setup is written similar to the usual controller test suite_test.go,
providing flexibility in the test setup and execution. The structure tries
to keep the infrastructure code separate from the test code such that
the infrastructure provisioning tool can be easily replaced, without the
need to change the actual tests.

The go tests are also written similar to the usual controller tests,
with a kube client from the tftestenv environment global variable.

Terraform is automatically downloaded using the terraform-exec package.
The test container images are generated using the go-containerregistry
package and pushed to the test container image repository.

`tests/integration/README.md` contains more details about the test setup and workflow.

The terraform configurations for EKS is based on https://github.com/terraform-aws-modules/terraform-aws-eks/tree/6bb73bbdd469797cb067e9aa22ce2f31a7fb22be/examples/self_managed_node_group .

The terraform configurations for AKS is based on https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster .

The terraform configurations for GKE is based on the official GKE terraform provider examples https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster with [Terraform Kubernetes Engine Auth Module](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/21.1.0/submodules/auth) to be able to easily retrieve the kubeconfig. Both GCR and Artifact Registry are tested.

Average test duration including provisioning and cleanup time:
- EKS - 25-30 minutes
- AKS - ~13 minutes
- GKE - ~9 minutes

We may be able to optimize some configurations to bring down the time of EKS and AKS in the future.

Part of  #264.

Issue #264, opened 2022-05-27
Extract auto-login code to fluxcd/pkg


stefanprodan:
As described in the RFC [Flux OCI support for Kubernetes manifests](https://github.com/fluxcd/flux2/pull/2601), source-controller will have to use the same flags and authentication logic for Azure, AWS and Google Cloud as implemented in this controller. 

We need to extract the auto-login code and the flags: `--aws-autologin-for-ecr`, `--azure-autologin-for-acr`, `--gcp-autologin-for-gcr` to a dedicated package in `fluxcd/pkg/oci/auth` so that we can share the code between image-reflector-controller and source-controller.

withnale:
I was searching for this issue on the source-controller but have found it here. Is there an eta on this?

The workload-identity logic works well in image-reflector, but I assume to get source-controller (and HelmRepository) able to use `oci:` then this card will need to be actioned and this logic be available in both apps?

PR #276, opened 2022-06-21
Introduce registry package

darkowlzz:
`internal/registry` package consolidates all the registry provider logins.
The `internal/registry/login` package contains a login `Manager` which
manages logins for all the providers. For testability, it provides methods to
modify the provider client configurations.
`internal/registry/{aws/azure/gcp}` packages contain clients for logging
into the respective registry. The client APIs are mostly similar across all the
providers, except for the small details related to overriding certain
configurations for testing purposes. Each of the providers have test
coverage to solidify the expected behavior in different scenarios.

This depends on #275 for testing to ensure that the native registry login
tests work before and after these changes.

NOTE: I'd like to gain some confidence in this package, maybe an IRC
release with this package, before moving it to `fluxcd/pkg` repo. It'd be
easier to fix any potential issues if it's in the same repo with all the
integration tests.

Part of  #264

### DIFF
--- a/internal/registry/aws/auth.go
+++ b/internal/registry/aws/auth.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/google/go-containerregistry/pkg/authn"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/fluxcd/image-reflector-controller/internal/registry"
+)
+
+var registryPartRe = regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
+
+// ParseImage returns the AWS account ID and region and `true` if
+// the image repository is hosted in AWS's Elastic Container Registry,
+// otherwise empty strings and `false`.
+func ParseImage(image string) (accountId, awsEcrRegion string, ok bool) {
+	registryParts := registryPartRe.FindAllStringSubmatch(image, -1)
+	if len(registryParts) < 1 || len(registryParts[0]) < 3 {
+		return "", "", false
+	}
+	return registryParts[0][1], registryParts[0][2], true
+}
+
+// Client is a AWS ECR client which can log into the registry and return
+// authorization information.
+type Client struct {
+	*aws.Config
+}
+
+// NewClient creates a new ECR client with default configurations.
+func NewClient() *Client {
+	return &Client{Config: aws.NewConfig()}
+}
+
+// getLoginAuth obtains authentication for ECR given the account
+// ID and region (taken from the image). This assumes that the pod has
+// IAM permissions to get an authentication token, which will usually
+// be the case if it's running in EKS, and may need additional setup
+// otherwise (visit
+// https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ as a
+// starting point).
+func (c *Client) getLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error) {
+	// No caching of tokens is attempted; the quota for getting an
+	// auth token is high enough that getting a token every time you
+	// scan an image is viable for O(500) images per region. See
+	// https://docs.aws.amazon.com/general/latest/gr/ecr.html.
+	var authConfig authn.AuthConfig
+	accountIDs := []string{accountId}
+
+	// Configure session.
+	cfg := c.Config.WithRegion(awsEcrRegion)
+	ecrService := ecr.New(session.Must(session.NewSession(cfg)))
+	ecrToken, err := ecrService.GetAuthorizationToken(&ecr.GetAuthorizationTokenInput{
+		RegistryIds: aws.StringSlice(accountIDs),
+	})
+	if err != nil {
+		return authConfig, err
+	}
+
+	// Validate the authorization data.
+	if len(ecrToken.AuthorizationData) == 0 {
+		return authConfig, errors.New("no authorization data")
+	}
+	if ecrToken.AuthorizationData[0].AuthorizationToken == nil {
+		return authConfig, fmt.Errorf("no authorization token")
+	}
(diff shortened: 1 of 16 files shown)


internal/registry/aws/auth.go
```diff
@@ -0,0 +1,114 @@
...
+// ParseImage returns the AWS account ID and region and `true` if
+// the image repository is hosted in AWS's Elastic Container Registry,
+// otherwise empty strings and `false`.
+func ParseImage(image string) (accountId, awsEcrRegion string, ok bool) {
+	registryPartRe := regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
+	registryParts := registryPartRe.FindAllStringSubmatch(image, -1)
+	if len(registryParts) < 1 {
+		return "", "", false
+	}
+	return registryParts[0][1], registryParts[0][2], true
```
pjbgf, line 29: May lead to index out of bounds:

```suggestion
	if len(registryParts) < 3 {
		return "", "", false
	}
	return registryParts[0][1], registryParts[0][2], true
```

internal/registry/aws/auth.go
```diff
@@ -0,0 +1,114 @@
...
+// IAM permissions to get an authentication token, which will usually
+// be the case if it's running in EKS, and may need additional setup
+// otherwise (visit
+// https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ as a
+// starting point).
+func (c *Client) getLoginAuth(accountId, awsEcrRegion string) (authn.AuthConfig, error) {
+	// No caching of tokens is attempted; the quota for getting an
+	// auth token is high enough that getting a token every time you
+	// scan an image is viable for O(1000) images per region. See
+	// https://docs.aws.amazon.com/general/latest/gr/ecr.html.
```
pjbgf, line 54: I am not sure I follow the comment. From that page the `The maximum number of GetAuthorizationToken requests that you can make per second in the current Region` is 500.

internal/registry/aws/auth.go
```diff
@@ -0,0 +1,114 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/fluxcd/image-reflector-controller/internal/registry"
+)
+
+// ParseImage returns the AWS account ID and region and `true` if
+// the image repository is hosted in AWS's Elastic Container Registry,
+// otherwise empty strings and `false`.
+func ParseImage(image string) (accountId, awsEcrRegion string, ok bool) {
+	registryPartRe := regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
```
pjbgf, line 24: Ideally `regexp.MustCompile` would be placed outside of the func, so we can cache that operation. The performance impact:
```
BenchmarkCurrent-16               173902              5948 ns/op            9885 B/op         61 allocs/op
BenchmarkPreCompiled-16          1546942               832.4 ns/op           433 B/op          3 allocs/op
```

Code:
```golang
func BenchmarkCurrent(b *testing.B) {
	for i := 0; i < b.N; i++ {
		registryPartRe := regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
		_ = registryPartRe.FindAllStringSubmatch("012345678901.dkr.ecr.us-east-1.amazonaws.com/foo:v1", -1)
	}
}

func BenchmarkPreCompiled(b *testing.B) {
	registryPartRe := regexp.MustCompile(`([0-9+]*).dkr.ecr.([^/.]*)\.(amazonaws\.com[.cn]*)/([^:]+):?(.*)`)
	for i := 0; i < b.N; i++ {
		_ = registryPartRe.FindAllStringSubmatch("012345678901.dkr.ecr.us-east-1.amazonaws.com/foo:v1", -1)
	}
}
```

internal/registry/azure/auth.go
```diff
@@ -0,0 +1,110 @@
...
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
+		if err != nil {
+			return authConfig, err
+		}
+		c.credential = cred
+	}
+
+	// Obtain access token using the token credential.
+	armToken, err := c.credential.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{string(arm.AzurePublicCloud) + ".default"},
```
pjbgf, line 61: Is this only supported in public cloud? We can always extend this later, but it may be worth adding a comment to this effect.

internal/registry/azure/exchanger.go
```diff
@@ -63,48 +64,55 @@ type acrError struct {
...
 		return "", fmt.Errorf("failed to send token exchange request: %w", err)
 	}
 
 	if resp.StatusCode != http.StatusOK {
+		// Parse the error response.
 		var errors []acrError
 		decoder := json.NewDecoder(resp.Body)
 		if err = decoder.Decode(&errors); err == nil {
-			return "", fmt.Errorf("unexpected status code %d from exchnage request: errors:%s",
+			return "", fmt.Errorf("unexpected status code %d from exchange request: errors:%s",
```
pjbgf, line 104: We are missing an space here. But potentially `errors:` is redundant and could be completely removed.

```suggestion
			return "", fmt.Errorf("unexpected status code %d from exchange request: errors: %s",
```

darkowlzz:
@pjbgf thanks for reviewing the actual login code. Since we didn't had any automated tests before for the cloud providers, I tried to keep the existing login code as they were and wrap them in a new structure, just to ensure I don't break anything. Now that we have automated tests, I'll go through your comments and address based on actual testing.

internal/registry/aws/auth.go
darkowlzz: Looks like this comment is based on flux v1 https://github.com/fluxcd/image-reflector-controller/commit/0ea1627bb8fe5302e3b56c2eb99b6c961da6e739.
It's possible that the numbers have changed now and we can change it to 500.

internal/registry/constants.go
```diff
@@ -0,0 +1,12 @@
+package registry
+
+// Provider is used to categorize the registry providers.
+type Provider int
+
+// Registry providers.
+const (
+	ProviderGeneric Provider = iota
```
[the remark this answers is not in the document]
darkowlzz: I've added a comment above it to clarify that it's used to just categorize the unsupported providers.
If it helps, the constant can also be renamed to `ProviderUnsupported`.

internal/registry/azure/exchanger.go
darkowlzz: Removed "errors:".

internal/registry/azure/auth.go
darkowlzz: Added a TODO for this.

internal/registry/aws/auth.go
darkowlzz: Moved the regexp outside, package global.

internal/registry/aws/auth.go
darkowlzz: Just tried it and looks like it was correct for ensuring `registryParts` is not zero, but the other indices are slices inside that slice, multi-dimensional. We can add another check for that, to make it safer.

darkowlzz: Just added one more check:
```diff
-       if len(registryParts) < 3 {
+       if len(registryParts) < 1 || len(registryParts[0]) < 3 {
```
Which would make sure it fails and not panics.

darkowlzz:
I've made some changes based on the review comments and tested it using the test automation.

stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

pjbgf (approved):
fantastik @darkowlzz! 🔝🔝 

LGTM

hiddeco (approved):
LGTM. Merci @darkowlzz :bow:

PR #287, opened 2022-07-21
cloud-provider-e2e: Use test image-reflector build

darkowlzz:
Update the cloud provider e2e test suite to take the image-reflector-controller
build to be tested using the variable `TEST_IMG`. This test image must be
built locally before running the e2e tests. In the e2e tests, this local image
is re-tagged and pushed into a remote registry and is used in flux deployment.

For Google Artifact Registry and Azure Container Registry, the
previously created test registries are reused to push the
image-reflector image. But for ECR, a new repository is created due to
lack of support for dynamic repositories.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -38,10 +38,16 @@ deleted manually.
 Copy `.env.sample` to `.env`, put the respective provider configurations in the
 environment variables and source it, `source .env`.
 
-Run the test with `make test-*`:
+Ensure the image-reflector-controller container image to be tested is built and
+ready for testing. A development image can be built from the root of the project
+by running the make target `docker-build`. Or, a release image can also be
+downloaded and used for testing.
+
+Run the test with `make test-*`, setting the image-reflector image, built or
+downloaded, with variable `TEST_IMG`:
 
 ```console
-$ make test-aws
+$ make test-aws TEST_IMG=foo/image-reflector-controller:dev
 mkdir -p build/flux
 curl -Lo build/flux/install.yaml https://github.com/fluxcd/flux2/releases/latest/download/install.yaml
   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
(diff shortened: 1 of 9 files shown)


tests/integration/README.md
```diff
@@ -38,10 +38,14 @@ deleted manually.
 Copy `.env.sample` to `.env`, put the respective provider configurations in the
 environment variables and source it, `source .env`.
 
-Run the test with `make test-*`:
+Ensure the image-reflector-controller container image to be tested is built and
+ready for testing.
+
```
somtochiama, line 43: Should we add a small note here on how to build the `image-controller-image`?

darkowlzz: I think a brief pointer to the Makefile at the root of the repo would be good.

I was also thinking of adding a check if the given `TEST_IMG` exists in the local docker image store and if not, fail. If we do that, I think then we can put a default value for `TEST_IMG`, and print a message that it doesn't exist when it's not explicitly set.

Sure, we can also add some instructions about testing a stable release image.

tests/integration/README.md
```diff
@@ -38,10 +38,14 @@ deleted manually.
 Copy `.env.sample` to `.env`, put the respective provider configurations in the
 environment variables and source it, `source .env`.
```
somtochiama, line 40: The different variables for each provider in `.env.sample` are commented out.
We can remind users here to uncomment it (or change the `.env.sample` so that they aren't commented).

tests/integration/README.md
somtochiama: > I was also thinking of adding a check if the given TEST_IMG exists in the local docker image store and if not, fail.

This is a good idea. Right now, it fails after the terraform infrastructure has already been created. We can add the check much earlier.

tests/integration/README.md
darkowlzz: > put the respective provider configurations in the environment variables and source it

should mean the same as it's talking about sourcing env vars.

tests/integration/README.md
darkowlzz: Added a docker image inspect command to check if the image already exists.

pjbgf (approved):
LGTM

Issue #289, opened 2022-07-26
image-reflector-controller cannot pull image list


ccodreanu:
We upgraded from 0.19.2 to 0.19.3 and our Image Policies are getting spammed with:

```
image-reflector-controller  Cannot determine latest tag for policy: version list argument cannot be empty
```

while Image Repositories show:

```
image-reflector-controller  GCR authentication failed: registry provider not configured for login
```

We see this happening on images like `k8s.gcr.io/sig-storage/snapshot-validation-webhook` which do not require auth to be pulled.

Image Repo and Policy:

```
---
apiVersion: image.toolkit.fluxcd.io/v1beta1
kind: ImageRepository
metadata:
  name: snapshot-validation-webhook
spec:
  image: k8s.gcr.io/sig-storage/snapshot-validation-webhook
  interval: 24h
---
apiVersion: image.toolkit.fluxcd.io/v1beta1
kind: ImagePolicy
metadata:
  name: snapshot-validation-webhook
spec:
  imageRepositoryRef:
    name: snapshot-validation-webhook
  policy:
    semver:
      range: ">=5.0.0"
```

darkowlzz:
Hi, thanks for reporting the issue. We did some refactoring of the registry code in the last release and looks like we didn't had a coverage for this scenario. I've created #290 to fix this issue and also added a test for it so that we don't have this bug again in the future.

darkowlzz:
Image reflector controller v0.19.4 is available now https://github.com/fluxcd/image-reflector-controller/pkgs/container/image-reflector-controller/29796368?tag=v0.19.4 . You can give it a try.

ccodreanu:
Works, thanks!

PR #290, opened 2022-07-26
imagerepo: Continue scan for unconfigured provider

darkowlzz:
When a given image reference is identified as a known provider but is
unconfigured, continue the reconciliation and scan to be able to scan
public repositories. Abort the reconciliation for any other error during
login.

Also add a test for scanning public repos without rate limits.

Fixes https://github.com/fluxcd/image-reflector-controller/issues/289


stefanprodan (approved):
LGTM

Thanks @darkowlzz

PR #292, opened 2022-07-28
tests/integration: Use terraform modules and test-infra/tftestenv

darkowlzz:
Updates all the integration tests for the cloud providers to use
terraform modules. All the state output variables exposed to the tests
are kept the same, only the terraform configurations have been changed.

Use tftestenv from github.com/fluxcd/test-infra/tftestenv.


stefanprodan (approved):
LGTM 

Thanks @darkowlzz

PR #297, opened 2022-08-08
Add flags to configure exponential back-off retry

pjbgf:
Add two new flags to enable users to configure exponential back-off for Flux objects. The default values are now set to 750ms for minimum retry time, and 15min for max.

Fixes: https://github.com/fluxcd/image-reflector-controller/issues/246.


pjbgf:
Keeping this as draft as it under goes some long running tests.

darkowlzz (approved):
LGTM!

Issue #299, opened 2022-08-08
Controller errors with `version list argument cannot be empty` after being restarted


pjbgf:
At controller restart, the controller may error: `Cannot determine latest tag for policy: version list argument cannot be empty`:

`
{"level":"error","ts":"2022-08-08T13:49:44.789Z","logger":"controller.imagepolicy","msg":"Reconciler error","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImagePolicy","name":"my-image-policy","namespace":"flux-system","error":"Cannot determine latest tag for policy: version list argument cannot be empty"}
`

PR #300, opened 2022-08-08
Skip error policy reconciliation if no tags are found

pjbgf:
At controller restart, it is quite common to experience `Cannot determine latest tag for policy: version list argument cannot be empty` errors. This happens because the existing repository may be Ready, but all its underlying data has been lost as it resides in the controller. 

Instead of erroring, this PR tries to self-heal by postponing the Policy reconciliation based on its repository set interval.

As a result, the controller will log:
`
{"level":"info","ts":"2022-08-08T17:37:52.544Z","logger":"controller.imagepolicy","msg":"no tags found in 'my-image-policy', retrying in 1m0s","reconciler group":"image.toolkit.fluxcd.io","reconciler kind":"ImagePolicy","name":"source-controller","namespace":"flux-system"}
`

And generate a Kubernetes Event with similar information.

Fixes: https://github.com/fluxcd/image-reflector-controller/issues/299

### DIFF
--- a/controllers/imagepolicy_controller.go
+++ b/controllers/imagepolicy_controller.go
@@ -182,6 +182,14 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		var tags []string
 		tags, err = r.Database.Tags(repo.Status.CanonicalImageName)
 		if err == nil {
+			if len(tags) == 0 {
+				msg := fmt.Sprintf("no tags found in local storage for '%s'", repo.Name)
+				r.event(ctx, pol, events.EventSeverityInfo, msg)
+				log.Info(msg)
+
+				return ctrl.Result{}, nil
+			}
+
 			var filter *policy.RegexFilter
 			if pol.Spec.FilterTags != nil {
 				filter, err = policy.NewRegexFilter(pol.Spec.FilterTags.Pattern, pol.Spec.FilterTags.Extract)


controllers/imagepolicy_controller.go
```diff
@@ -180,6 +180,13 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		var tags []string
 		tags, err = r.Database.Tags(repo.Status.CanonicalImageName)
 		if err == nil {
+			if len(tags) == 0 {
+				msg := fmt.Sprintf("no tags found in '%s', retrying in %s", repo.Name, repo.Spec.Interval.Duration)
+				r.event(ctx, pol, events.EventSeverityInfo, msg)
```
stefanprodan, line 185: Please log this as well, issuing events only is not something we should do IMO.

pjbgf:
Test image as of https://github.com/fluxcd/image-reflector-controller/pull/300/commits/39cc9bf699bd57864e0e5c63bf03de1e49e6ded3: `ghcr.io/fluxcd/image-reflector-controller:rc-39cc9bf6`

controllers/imagepolicy_controller.go
pjbgf: Added the logs and will generate another rc image with it.

pjbgf:
Test image based on 3826a423: ghcr.io/fluxcd/image-reflector-controller:rc-3826a423

controllers/imagepolicy_controller.go
```diff
@@ -180,6 +180,14 @@ func (r *ImagePolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		var tags []string
 		tags, err = r.Database.Tags(repo.Status.CanonicalImageName)
 		if err == nil {
+			if len(tags) == 0 {
+				msg := fmt.Sprintf("no tags found in '%s', retrying in %s", repo.Name, repo.Spec.Interval.Duration)
```
stefanprodan, line 184: Maybe we should mention where we didn’t found them?
```suggestion
				msg := fmt.Sprintf("no tags found in local storage for '%s', retrying in %s", repo.Name, repo.Spec.Interval.Duration)
```

controllers/imagepolicy_controller.go
stefanprodan: Thanks

stefanprodan (approved):
LGTM

darkowlzz (approved):
LGTM! Thanks.

PR #307, opened 2022-09-02
fuzz: Fuzz optimisations

pjbgf:
General optimisations in preparation for converting to Go native fuzz:
- Ensures CI runs on Go 1.18.
- Fix caching path and reuse cache from host.
- Ensure latest oss-fuzz base images are being used.

Relates to https://github.com/fluxcd/flux2/issues/2417.

### DIFF
--- a/.github/workflows/cifuzz.yaml
+++ b/.github/workflows/cifuzz.yaml
@@ -13,12 +13,19 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+    - name: Setup Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.x
+    - id: go-env
+      run: |
+        echo "::set-output name=go-mod-cache::$(go env GOMODCACHE)"       
     - name: Restore Go cache
       uses: actions/cache@v3
       with:
-        path: /home/runner/work/_temp/_github_home/go/pkg/mod
+        path: ${{ steps.go-env.outputs.go-mod-cache }}
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
         restore-keys: |
-          ${{ runner.os }}-go-
+          ${{ runner.os }}-go
     - name: Smoke test Fuzzers
       run: make fuzz-smoketest
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ fuzz-build:
 	rm -rf $(shell pwd)/build/fuzz/
 	mkdir -p $(shell pwd)/build/fuzz/out/
 
-	docker build . --tag local-fuzzing:latest -f tests/fuzz/Dockerfile.builder
+	docker build . --pull --tag local-fuzzing:latest -f tests/fuzz/Dockerfile.builder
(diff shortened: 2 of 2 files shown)


.github/workflows/cifuzz.yaml
```diff
@@ -13,12 +13,16 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+    - name: Setup Go
```
relu, line 16: Since we're running inside a container do we actually need this?

.github/workflows/cifuzz.yaml
```diff
@@ -13,12 +13,16 @@ jobs:
...
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.x
     - name: Restore Go cache
       uses: actions/cache@v3
       with:
-        path: /home/runner/work/_temp/_github_home/go/pkg/mod
+        path: ~/go/pkg/mod
```
relu, line 23: I think for clarity it would still be a good idea to use the absolute path explicitly or rely on `GOMODCACHE` which is used in the makefile target to mount the volume.

Makefile
```diff
@@ -139,10 +139,11 @@ fuzz-build:
 	rm -rf $(shell pwd)/build/fuzz/
 	mkdir -p $(shell pwd)/build/fuzz/out/
 
-	docker build . --tag local-fuzzing:latest -f tests/fuzz/Dockerfile.builder
+	docker build . --pull --tag local-fuzzing:latest -f tests/fuzz/Dockerfile.builder
 	docker run --rm \
```
relu, line 143: By switching to native fuzz I'm wondering do we still need this image?

pjbgf: Great question! When this project moves to native fuzz we will need to keep this or find a better way to ensure we are not breaking the [oss-fuzz](https://github.com/google/oss-fuzz) builds upstream.

One of the reasons we use our approach is to have a "fail-fast" repository-specific test. We could drop this off and rely on the [ci-fuzz action](https://google.github.io/oss-fuzz/getting-started/continuous-integration/), which will assess [all our repositories instead](https://github.com/google/oss-fuzz/blob/master/projects/fluxcd/Dockerfile#L25-L26).

I think we can reconsider that once we have all repos migrated to native fuzz. The only issue is that it will be a steeper learning curve for folks to find issues and mitigate them, specially when they work for native fuzz, but break on oss-fuzz. Another point to take into account will be the time taken on cifuzz and the flakiness it may bring, as I am not sure we will be able to make it repo-specific.

.github/workflows/cifuzz.yaml
pjbgf: Without Go we would not have `go env GOMODCACHE`. I leaned on this for consistency as well, given that other `fluxcd` repos require this due to [harder dependencies](https://github.com/fluxcd/source-controller/blob/6d479e559babb6399ff0b6f8a4fc336eb0d95421/Makefile#L183).

stefanprodan (approved):

relu (approved):

Issue #166, opened 2021-08-24
Limit scanning success events to be less spammy


squaremo:
As reported in https://github.com/fluxcd/image-reflector-controller/pull/165#discussion_r694677790, sending an event on every successful scan is considered too spammy. Consider e.g., limiting those events to just the scans that discover new tags.

marioapardo:
The other thing is that the event sends a text that there is no new image and thus be able to create an exclusion.

PR #311, opened 2022-09-13
Refactor reconcilers and introduce v1beta2 API

darkowlzz:
New status conditions in various scenarios:
- ImageRepository - https://gist.github.com/darkowlzz/19b0ef234f1c68baa42365726db1028e
- ImagePolicy - https://gist.github.com/darkowlzz/bb017ba22a932fd4e7e2a4f421dbcbd2

Uses the result finalizer from https://github.com/fluxcd/pkg/pull/329

TODO: Add more details about the change.


stefanprodan:
@darkowlzz please delete the v1alpha CRDs, we haven't kept them around for the others controllers, so it's time to remove them here too.

stefanprodan:
For `ImageRepository` In propose that we add the latest ten tags fetched from the registry. The tag list would help users debug polices by looking up `.status.lastScanResult.latestTags` e.g.:

```yaml
status:
  canonicalImageName: ghcr.io/stefanprodan/podinfo
  conditions:
  - lastTransitionTime: "2022-09-12T11:59:49Z"
    message: successful scan, found 34 tags
    observedGeneration: 1
    reason: Succeeded
    status: "True"
    type: Ready
  lastScanResult:
    scanTime: "2022-09-12T11:59:49Z"
    tagCount: 34
    latestTags: '6.2.0, 6.1.8, 6.1.7, 6.1.6, 6.1.5, 6.1.4, 6.1.3, 6.1.2, 6.1.1, 6.1.0, ...'
  observedExclusionList:
  - ^.*\.sig$
  observedGeneration: 1
```

stefanprodan:
@darkowlzz does this PR solves https://github.com/fluxcd/image-reflector-controller/issues/166?

darkowlzz:
> does this PR solves #166?

@stefanprodan yes. In https://github.com/fluxcd/image-reflector-controller/blob/5b6015b03766c99309a3418838c0df07b7aa3800/controllers/imagerepository_controller.go#L185-L188 it tries to only send notification when the result changes. I'll do some more testing to make sure it's handled well.

darkowlzz:
Added a rough implementation of latest tags in ImageRepository status:

```yaml
status:
  canonicalImageName: ghcr.io/stefanprodan/podinfo
  conditions:
  - lastTransitionTime: "2022-09-13T23:23:46Z"
    message: successful scan, found 34 tags
    observedGeneration: 1
    reason: Succeeded
    status: "True"
    type: Ready
  lastScanResult:
    latestTags: latest, 6.2.0, 6.1.8, 6.1.7, 6.1.6, 6.1.5, 6.1.4, 6.1.3, 6.1.2, 6.1.1
      ...
    scanTime: "2022-09-13T23:48:52Z"
    tagCount: 34
  observedExclusionList:
  - ^.*\.sig$
  observedGeneration: 1
```
Looks like we may have to remove `latest` from it.

Also added previous tag in the new ImagePolicy version events.
When it's first policy apply, it results in event:
```
Latest image tag for 'ghcr.io/stefanprodan/podinfo' resolved to: 5.2.0
```
When there's an update, it results in event:
```
Latest image tag for 'ghcr.io/stefanprodan/podinfo' updated from 5.2.0 to 5.2.1
```
This requires introducing a new field in ImagePolicy, `observedPreviousImage` to keep track of the previous image in order to prevent duplicate events on subsequent reconciliation without any actual update.

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-09-13T23:03:29Z"
    message: Latest image tag for 'ghcr.io/stefanprodan/podinfo' updated from 5.2.0
      to 5.2.1
    observedGeneration: 1
    reason: Succeeded
    status: "True"
    type: Ready
  latestImage: ghcr.io/stefanprodan/podinfo:5.2.1
  observedGeneration: 1
  observedPreviousImage: ghcr.io/stefanprodan/podinfo:5.2.0
```

These are implemented in the last commit. Need some feedback and help in improving the messaging.
I still need to test them more and write some more unit tests.

stefanprodan:
> Looks like we may have to remove latest from it.

No we don't, we just trim the list to max 10 items without altering them.

> Need some feedback and help in improving the messaging.

I suggest we do:

```
Latest image tag for 'ghcr.io/stefanprodan/podinfo' resolved to 5.2.0

Latest image tag for 'ghcr.io/stefanprodan/podinfo' updated from 5.2.0 to 5.2.1
```

stefanprodan:
@tun0 you would setup alerting for `ImagePolicy` objects, and those will tell you "Latest image tag for 'ghcr.io/stefanprodan/podinfo' updated from 5.2.0 to 5.2.1", then the `ImageAutomation` alert will tell you "Image 'ghcr.io/stefanprodan/podinfo' updated from 5.2.1".

tun0:
The `ImagePolicy` stuff doesn't have any workload specific information though. The policy could've gone from 5.2.0 to 5.2.1, which won't say anything about a workload that was still on 5.1.0 for some reason. Perhaps it's a new resource that was just added to git with an outdated tag. Ideally I'd be able to see detailed version transitions per workload, not per policy. In the end, the people looking these notifications, in our case at least, care about workloads, not image policies.

stefanprodan:
OK so that issue is wrongly placed here, @darkowlzz please move it or create a new one in IAC.

darkowlzz:
Okay, my bad. I moved it to image-reflector yesterday because of the misunderstanding. I'll move it to IAC again and add some more context around it. Thanks @tun0 for the clarification.

stefanprodan:
> successful scan, found 33 tags

What happens at the next scan if the same tags are fetched, do we still issue an event or do we log only?

darkowlzz:
> What happens at the next scan if the same tags are fetched, do we still issue an event or do we log only?

Then it logs and emits k8s only event: `successful scan, next scan in 5m0s` .

stefanprodan:
> Then it logs and emits k8s only event: successful scan, next scan in 5m0s .

Let's make that "no new tags found, next scan in 5m"

darkowlzz:
The code should be ready for review now while I'm still working on the spec docs.
https://github.com/fluxcd/pkg/pull/329 should be ready in its current state. I'll address the remaining comments before marking the PR ready.

Regarding the ImageRepository events discussed above, now it only emits single event per reconciliation with different messages.
When a new repository is created and it scans successfully:
```console
successful scan, found 211 tags
```
This is also sent to the notification controller.

On subsequent reconciliation at the scan interval with no new tags:
```console
no new tags found, next scan in 1h0m0s
```
This is not sent to the notification controller. It's only logged and recorded as k8s event.

On reconciliation triggered before the scan interval without any configuration change that affect the scan result:
```console
no change in repository configuration since last scan, next scan in 58m39.516114621s
```
This is also not sent to the notification controller. It's only logged and recorded as k8s event.

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,610 @@
...
+These will often go together in case of self-hosted image registry. All the
+files in the secret are expected to be [PEM-encoded][pem-encoding]. This is an
+ASCII format for certificates and keys; `openssl` and such tools typically
+provide an option of PEM output.
+
+Assuming that a certificate file and private key are in files `client.crt` and
+`client.key` respectively, a secret can be created with `kubectl`:
+
+```sh
+SECRET_NAME=tls-certs
```
stefanprodan, line 173: Please remove the bash variable from this section.

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,610 @@
...
+  image: docker.io/org/image
+  exclusionList:
+    - "^.*\\.sig$"
+    - "1.0.2"
+    - "1.1.1|1.0.0"
+```
+
+## Working with ImageRepositories
+
+### Contextual Authentication
```
stefanprodan, line 286: I'm for holding off the `v1beta2` release until we add the `provider` field to the spec and we deprecate the controller flags. This is not blocking this PR but rather the release of the new API version.

stefanprodan (changes requested):
Please add validation to all `v1.Duration` fields, example here: https://github.com/fluxcd/source-controller/pull/903

docs/spec/v1beta2/imagerepositories.md
darkowlzz: `provider` is now added in the ImageRepository API and the respective integration tests have been updated to test it against all the cloud providers. Ran the tests myself and they work.

darkowlzz:
>  At present, I've removed the autologin flags as well. I'm not sure if we should allow existing users to experience failure due to unknown flags or we should keep the flags, not do anything with it, and remove them later.

After thinking more about it, I think we need to fail as this change requires updating the existing ImageRepository objects and not failing to run image-reflector-controller would eventually result in image scan to fail, so errors and user intervention are inevitable.

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
...
+```yaml
+---
+apiVersion: image.toolkit.fluxcd.io/v1beta2
+kind: ImageRepository
+metadata:
+  name: podinfo
+  namespace: default
+spec:
+  image: stefanprodan/podinfo
+  interval: 1h
```
stefanprodan, line 20: Let's add `provider: generic` to this example and to the explanation below (linking to the provider section).

controllers/imagepolicy_controller.go
```diff
@@ -39,241 +40,378 @@ import (
...
-		if err := r.patchStatus(ctx, req, pol.Status); err != nil {
-			return ctrl.Result{Requeue: true}, err
+	// Write the observations on status.
+	obj.Status.LatestImage = repo.Spec.Image + ":" + latest
+	// If the old latest image and new latest image don't match, set the old
+	// image as the observed previous image.
+	// NOTE: The following allows the previous image to be set empty when
+	// there's a failure and a subsequent recovery from it. This behavior helps
+	// avoid creating update event as there's no previous image to infer from.
+	// Recovery from a failure shouldn't result in update event.
```
pjbgf, line 309:
```suggestion
	// avoid creating an update event as there's no previous image to infer from.
	// Recovery from a failure shouldn't result in an update event.
```

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
+# Image Repositories
+
+The `ImageRepository` API defined a repository to scan and store a specific set
```
pjbgf, line 3:
```suggestion
The `ImageRepository` API defines a repository to scan and store a specific set
```

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
...
+### Image
+
+`.spec.image` is a required field that specifies the address of an image
+repository without any scheme prefix, e.g. `fluxcd/image-reflector-controller`.
+This image is converted to its canonical form by the controller before scanning.
+The canonical form of the image is reflected in `.status.canonicalImageName`.
+
+### Interval
+
+`.spec.interval` is a required field that specified the interval at which the
```
pjbgf, line 94:
```suggestion
`.spec.interval` is a required field that specifies the interval at which the
```

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
...
+Image repository must be scanned.
+
+After successfully reconciling the object, the image-reflector-controller reques
```
pjbgf, line 97:
```suggestion
After successfully reconciling the object, the image-reflector-controller re-queues
```

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
...
+
+`.spec.certSecretRef` is an optional field to specify a name reference to a
+Secret in the same namespace as the ImageRepository containing TLS certificate
+data. This is for two separate purposes:
+- to provide a client certificate and private key, if you use a certificate to authenticate with the image registry; and,
+- to provide a CA certificate, if the registry uses a self-signed certificate
+These will often go together in case of self-hosted image registry. All the
+files in the secret are expected to be [PEM-encoded][pem-encoding]. This is an
+ASCII format for certificates and keys; `openssl` and such tools typically
+provide an option of PEM output.
```
pjbgf, line 167: Maybe?
```suggestion
provide an option for PEM output.
```

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,723 @@
...
+    namespaceSelectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: flux-system
+```
+
+**Note** that the `kubernetes.io/metadata.name` is a readonly label added by
+Kubernetes >= 1.21 automatically on namespaces. For older version of Kubernetes,
+please set labels on the namespaces where the ImagePolicy exist.
+
+The above definition, allows or ImagePolicy in the `flux-system` namespace to
```
pjbgf, line 230:
```suggestion
The above definition, allows for ImagePolicy in the `flux-system` namespace to
```

pjbgf (commented):
A few nits

darkowlzz:
This is ready along with the spec docs. I've also updated the gists linked in the PR description that shows the status conditions in various scenarios.
While writing the spec docs, I realized that the Reconciling conditions can provide better information about the state when Ready=False, like what's the reason for a scan, new image, new exclusion list, scan interval, etc. Added some detailed Reconciling conditions in both ImageRepository and ImagePolicy for the same and they are described in the spec docs.

go.mod
```diff
@@ -7,12 +7,13 @@ replace github.com/fluxcd/image-reflector-controller/api => ./api
...
 	github.com/AdaLogics/go-fuzz-headers v0.0.0-20220903154154-e8044f6e4c72
 	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/darkowlzz/controller-check v0.0.0-20220902134353-b2ee0ae3566c
 	github.com/dgraph-io/badger/v3 v3.2103.2
 	github.com/fluxcd/image-reflector-controller/api v0.22.0
 	github.com/fluxcd/pkg/apis/acl v0.1.0
 	github.com/fluxcd/pkg/apis/meta v0.16.0
 	github.com/fluxcd/pkg/oci v0.10.0
-	github.com/fluxcd/pkg/runtime v0.19.0
+	github.com/fluxcd/pkg/runtime v0.19.1-0.20221004174449-09fc5a85dbff
```
stefanprodan, line 16: Please update runtime to latest stable and replace `controller-check`

controllers/imagerepository_controller.go
```diff
@@ -49,190 +44,251 @@ import (
...
+		log.Info("reconciliation is suspended for this object")
+		return ctrl.Result{}, nil
 	}
 
-	// If the object is under deletion, record the readiness, and remove our finalizer.
-	if !imageRepo.ObjectMeta.DeletionTimestamp.IsZero() {
-		r.recordReadinessMetric(ctx, &imageRepo)
-		controllerutil.RemoveFinalizer(&imageRepo, imagev1.ImageRepositoryFinalizer)
-		if err := r.Update(ctx, &imageRepo); err != nil {
-			return ctrl.Result{}, err
```
stefanprodan, line 132: I think that moving the suspend check before deletion, makes all suspended resources to hand forever on deletion, since the controller no longer removes the finalizer. Also SC has the same bug after refactoring cc @hiddeco

aryan9600:
was 10 chosen for a specific reason or is it just an arbitrary number?

api/v1beta2/imagerepository_types.go
```diff
@@ -0,0 +1,200 @@
...
+	// from being stored in the database.
+	// +kubebuilder:default:={"^.*\\.sig$"}
+	// +kubebuilder:validation:MaxItems:=25
+	// +optional
+	ExclusionList []string `json:"exclusionList,omitempty"`
+
+	// The provider used for authentication, can be 'aws', 'azure', 'gcp' or 'generic'.
+	// When not specified, defaults to 'generic'.
+	// +kubebuilder:validation:Enum=generic;aws;azure;gcp
+	// +kubebuilder:default:=generic
```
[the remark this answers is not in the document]
stefanprodan: The defaults only work at admission, so none of the existing objects will get them after conversion, only newly created objects get the defaults. IMO we shouldn't be using defaults if we care about backwards compatibility and we want users to be able to rollback. I very much regret having a default for `gitImplementation` which is now deprecated and no longer use, we are now in a situation where we can't delete it from the CRD because everyone has a value for it in etcd.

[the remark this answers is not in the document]
stefanprodan: > If you check my last comment above about upgrade testing, the defaults are added to existing objects.

I don't see that on my test cluster (v1.25.3).

Before upgrade:

```yaml
kind: ImageRepository
metadata:
  creationTimestamp: "2023-02-08T15:45:54Z"
  generation: 1
  name: podinfo
  namespace: default
  resourceVersion: "591"
  uid: 200bca96-f5d0-4ec3-bd75-0900d385c59b
spec:
  accessFrom:
    namespaceSelectors:
    - matchLabels:
        kubernetes.io/metadata.name: flux-system
  image: ghcr.io/stefanprodan/podinfo
  interval: 1m0s
status:
  observedGeneration: -1
```

After upgrade there no defaults:

```yaml
kind: ImageRepository
metadata:
  creationTimestamp: "2023-02-08T15:45:54Z"
  generation: 1
  name: podinfo
  namespace: default
  resourceVersion: "591"
  uid: 200bca96-f5d0-4ec3-bd75-0900d385c59b
spec:
  accessFrom:
    namespaceSelectors:
    - matchLabels:
        kubernetes.io/metadata.name: flux-system
  image: ghcr.io/stefanprodan/podinfo
  interval: 1m0s
status:
  observedGeneration: -1
```

But if I apply the object again, then the defaults are set because it happens at admission:

```yaml
kind: ImageRepository
metadata:
  creationTimestamp: "2023-02-08T15:45:54Z"
  generation: 1
  name: podinfo
  namespace: default
  resourceVersion: "1086"
  uid: 200bca96-f5d0-4ec3-bd75-0900d385c59b
spec:
  accessFrom:
    namespaceSelectors:
    - matchLabels:
        kubernetes.io/metadata.name: flux-system
  exclusionList:
  - ^.*\.sig$
  image: ghcr.io/stefanprodan/podinfo
  interval: 1m0s
  provider: generic
status:
  observedGeneration: -1
```

tun0:
> * Introduce observed previous image in ImagePolicy status to keep track of the previous image so that it can provide update path information.

The way I read this, is that this is about latest from the point-of-view of the policy. Our existing Flux v1 notifications show the "actual" previous image, as in: what was actually configured in git for the specific workload. 

The difference is subtle, but especially for new workloads it could be confusing:
* previousTag: foo
* newTag: bar
* create new workload X with "ancient" tag: qux
This would trigger a notification for X mentioning updating from foo to bar, whereas it should actually say from qux to bar.

I understand this is a bit of a corner-case, but on the other hand I also kinda treat this as feature parity with Flux v1.

tun0:
> > * create new workload X with "ancient" tag: qux
> >   This would trigger a notification for X mentioning updating from foo to bar, whereas it should actually say from qux to bar.
> 
> @tun0 Sounds like a confusion we had before, earlier in the thread. This notification is only for the image policy, independent of the workload. We'll add similar notification in image-automation-controller that'll be specific to the workload and would show the previously used tag that was in git and the new tag.

Ah, right. Makes sense!

api/v1beta2/imagerepository_types.go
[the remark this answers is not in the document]
stefanprodan: I can't reproduce the dry-run error with Kubernetes Kind, what people reported for NC was specific to EKS. So we can leave the defaults in and deal with EKS bugs later.

docs/spec/v1beta2/imagerepositories.md
```diff
@@ -0,0 +1,751 @@
...
+```
+
+#### Trace emitted Events
+
+To view events for specific ImageRepository(s), `kubectl get events` can be used
+in combination with `--field-selector` to list the Events for specific objects.
+For example, running
+
+```sh
+kubectl get events --field-selector involvedObject.kind=ImageRepository,involvedObject.name=<repository-name>
```
stefanprodan, line 579:
```suggestion
kubectl events --for ImageRepository/<repository-name>
```

Let's use the new events command that orders events correctly.

stefanprodan (approved):
LGTM

Thanks @darkowlzz 🥇

somtochiama (approved):
LGTM!

I tested this on GKE with GCR and Artifact Registry and it worked fine.
Thank you @darkowlzz 🌹

tun0:
> We'll add similar notification in image-automation-controller that'll be specific to the workload and would show the previously used tag that was in git and the new tag.

What's the status of this?

tun0:
@darkowlzz: Thanks for the update. Looking forward to the request :wink: I'll gladly assist where possible.

PR #317, opened 2022-09-27
Dockerfile: Build with Go 1.19

stefanprodan:
Bump build image to Go 1.19

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,23 @@
+ARG GO_VERSION=1.19
 ARG XX_VERSION=1.1.0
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -29,7 +27,7 @@ COPY main.go main.go
 COPY controllers/ controllers/
 COPY internal/ internal/
 
-# build without giving the arch, so that it gets it from the machine
+# build
```
hiddeco, line 30: Probably good to add `-trimpath` to `xx-go build` now that we're making changes to the file anyway.

hiddeco (approved):

PR #323, opened 2022-11-09
Remove deprecated alpha APIs

stefanprodan:
Remove v1alpha1 and v1alpha2 APIs. These versions were deprecated more than a year ago, I think it's time to drop support for them as we did for all the other alpha APIs in Flux.


hiddeco (approved):

darkowlzz (commented):
Based on https://github.com/fluxcd/image-reflector-controller/pull/311/commits/cf6339d83d62f60db8b47560b43836eb2a9cdfb4 , looks like PROJECT file needs to be updated and some files under config/samples can also be removed.

makkes (approved):

pjbgf (approved):

PR #353, opened 2023-03-08
Fallback to autlogin flags if no provider is specified

### LINKED ISSUE #352: image-reflector-controller 0.40.0 breaking change gives no room for upgrading transparently
Problem: the `--aws-autologin-for-ecr` flag was deprecated and its effect removed at the same time. Now there’s no room for upgrading without breaking stuff.

Background: We have 100+ image reflections running, spread over a number of teams with their config in their own team repositories. All of these are using private AWS ECR, and we depend on the image-reflector-controller's `--aws-autologin-for-ecr` flag for authentication.

As soon as we upgrade Flux to 0.40+, the image-reflector-controller stops authenticating with ECR, and all teams’ image automations immediately stops working. Upgrading all ImageRepositories to v1beta2 would fix the problem, but it’s hard to coordinate this. 

I’d expect the image-reflector-controller to keep the flag and keep the behavior for `image.toolkit.fluxcd.io/v1beta1` API, allowing us to upgrade manifests to `v1beta2`, before removing the flag in a future release.

[Slack thread](https://cloud-native.slack.com/archives/CLAJ40HV3/p1678184516852989)

aryan9600:
Use the value of autologin flags if `spec.provider` is not set or set to `generic`.

Fixes #352

### DIFF
--- a/controllers/imagerepository_controller.go
+++ b/controllers/imagerepository_controller.go
@@ -114,6 +114,7 @@ type ImageRepositoryReconciler struct {
 		DatabaseWriter
 		DatabaseReader
 	}
+	DeprecatedLoginOpts login.ProviderOptions
 
 	patchOptions []patch.Option
 }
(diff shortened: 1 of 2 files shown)


hiddeco (approved):
With impact full changes like this (or reverts because we did something wrong). It's often good to include reasoning with your commit to make your case.

Issue #237, opened 2022-03-18
Add support for Azure Workload Identity


pjbgf:
Azure Workload Identity is the next generation of workload contextual authentication, which [replaces the existing Azure Pod Identity](https://cloudblogs.microsoft.com/opensource/2022/01/18/announcing-azure-active-directory-azure-ad-workload-identity-for-kubernetes/).

For more information: https://azure.github.io/azure-workload-identity/docs/introduction.html

somtochiama:
This was fixed by https://github.com/fluxcd/image-reflector-controller/pull/363

PR #361, opened 2023-03-21
Enable Azure workload identity

somtochiama:
Depends on: https://github.com/fluxcd/pkg/pull/450

Closes: #237


somtochiama:
This pull request was converted to a draft and the updates to dependencies were moved to a different pull request. Updates to the tests require more consideration

PR #365, opened 2023-03-29
Add reconciler sharding capability based on label selector

### LINKED ISSUE #364: Add reconciler sharding capability based on label selector
Part of fluxcd/flux2#3731

makkes:
The new `--watch-label-selector` lets operators provide a label to the controller manager which in turn uses it to reconcile only those resources (`ImageRepositories` and `ImagePolicies`) that match the given label.

This way operators can deploy multiple instances of IRC, each reconciling a distinct set of resources based on their labels and effectively scale IRC horizontally.

## ⚠️ WARNING

If sharding is enabled, all `ImagePolicy` resources can only refer to those `ImageRepository` resources that are captured by the exact same label selector as the `ImagePolicies`, otherwise the `ImagePolicies` will not become ready.

## Example

Given two Deployments with different shard keys:

1. Deployment `image-reflector-controller-1` with `--watch-label-selector=sharding.fluxcd.io/shard=shard1`
1. Deployment `image-reflector-controller-2` with `--watch-label-selector=sharding.fluxcd.io/shard=shard2`

An `ImageRepository` like this:

```yaml
kind: ImageRepository
metadata:
  labels:
    sharding.fluxcd.io/shard: shard1
[...]
```

An `ImagePolicy` like this:

```yaml
kind: ImagePolicy
metadata:
  labels:
    sharding.fluxcd.io/shard: shard1
[...]
spec:
  imageRepositoryRef:
    name: podinfo
[...]
```

The `ImagePolicy` will successfully be reconciled.

If the `ImagePolicy` would have the label `sharding.fluxcd.io/shard: shard2` reconciliation would fail because it would be referring to an `ImageRepository` from a different shard.

closes #364


stefanprodan:
For IRC things are a more complicated, can you please explain in the PR description that the same shard hey must be used for all policies referring to an repo including that repo.

go.mod
```diff
@@ -11,20 +11,20 @@ require (
...
 	github.com/spf13/pflag v1.0.5
 	go.uber.org/zap v1.24.0
-	k8s.io/api v0.26.2
-	k8s.io/apimachinery v0.26.2
-	k8s.io/client-go v0.26.2
-	sigs.k8s.io/controller-runtime v0.14.5
+	k8s.io/api v0.26.3
+	k8s.io/apimachinery v0.26.3
+	k8s.io/client-go v0.26.3
+	sigs.k8s.io/controller-runtime v0.14.6
```
stefanprodan, line 27: Please bump these in the API package.

stefanprodan (approved):
LGTM

Thanks @makkes

PR #372, opened 2023-05-08
Update dependencies

stefanprodan:
Packages:
- github.com/Masterminds/semver/v3 v3.2.1
- github.com/fluxcd/pkg/oci v0.23.0
- github.com/google/go-containerregistry v0.15.1


makkes (approved):
thanks

PR #387, opened 2023-05-30
Update integration tests and run them against Azure and GCP

darkowlzz:
- Update the integration tests to perform graceful stop in CI on interruption.
- Update terraform configs to use tags.
- Add instructions for using service accounts in Azure and GCP.
- Add workflows for running the tests against Azure and GCP.
- Increase test wait result timeout to 30s, fails with 20s sometimes.
- Update the way principal ID is passed to ACR, refer https://github.com/fluxcd/test-infra/pull/5 .

Changes are similar to the OCI integration test in https://github.com/fluxcd/pkg/pull/535.
Workflows are similar to the OCI test workflows added in https://github.com/fluxcd/pkg/pull/543.

__NOTE:__ The workflows will run once every day. Cleanup of the leftover resources will be taken care of by the cleanup workflow in pkg repo.

Example runs:
- Azure - https://github.com/darkowlzz/image-reflector-controller/actions/runs/5127236485/jobs/9222642446
- GCP - https://github.com/darkowlzz/image-reflector-controller/actions/runs/5127266689/jobs/9222696011

### DIFF
--- a/tests/integration/terraform/azure/main.tf
+++ b/tests/integration/terraform/azure/main.tf
@@ -25,7 +25,7 @@ module "acr" {
 
   name             = local.name
   location         = var.azure_location
-  aks_principal_id = module.aks.principal_id
+  aks_principal_id = [module.aks.principal_id]
   resource_group   = module.aks.resource_group
   tags             = var.tags
 


tests/integration/suite_test.go
```diff
@@ -177,6 +181,22 @@ func TestMain(m *testing.M) {
...
+
+		// Calling exit on panic prevents logging of panic error.
+		// Exit only on normal return. Explicitly detect panic and log the error
+		// on panic.
+		if err := recover(); err == nil {
+			os.Exit(exitCode)
+		} else {
+			log.Printf("panic: %v", err)
+		}
+	}()
```
stefanprodan, line 198: This looks great, we'll need this in all controllers.

stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

PR #423, opened 2023-07-21
tests/int: IAM CI setup docs & exit code propagation

darkowlzz:
Add terraform config in the docs for creating IAM permissions and populating github actions secrets and variables required by the integration tests.

Also, update the test runner to propagate the exit code always. This is based on observations while testing flux2 e2e tests that the exit code didn't surface in certain situations.

Similar to https://github.com/fluxcd/pkg/pull/607.


somtochiama (approved):
LGTM!

Thanks Sunny

Issue #479, opened 2023-12-08
bug(helm/aws): unknown build error: failed to get credential from 'aws': not found, ResolveEndpointV2


cep21:
My version tags:
ghcr.io/fluxcd/image-reflector-controller v0.31.0 
ghcr.io/fluxcd/source-controller v1.2.1

This is likely the cause: https://github.com/aws/aws-sdk-go-v2/issues/2370

Relevant quote:

> Please upgrade ALL of your dependencies under the github.com/aws/aws-sdk-go-v2 namespace to the latest tags from yesterday's release, that should resolve the issue. Please chime in with the results if you can.
> 
> It appears that due to the breaking changes around middleware reordering, service client modules tagged on or after yesterday are now fundamentally incompatible with runtime modules tagged before yesterday (and vice versa).
> 

It's possible someone did not update *every* sdk dependency

More:

Relevant status from the `HelmChart` object

```
    Last Transition Time:  2023-12-08T22:36:31Z
    Message:               unknown build error: failed to get credential from 'aws': not found, ResolveEndpointV2
    Observed Generation:   3
    Reason:                Unknown
    Status:                False
    Type:                  Ready
```

PR #481, opened 2023-12-11
Update dependencies

hiddeco:
Releated to https://github.com/fluxcd/image-reflector-controller/issues/479


makkes (approved):

api/go.mod
```diff
@@ -20,13 +20,13 @@ require (
...
 	github.com/modern-go/reflect2 v1.0.2 // indirect
-	golang.org/x/net v0.18.0 // indirect
+	golang.org/x/net v0.19.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/klog/v2 v2.100.1 // indirect
-	k8s.io/utils v0.0.0-20230726121419-3b25d923346b // indirect
+	k8s.io/klog/v2 v2.110.1 // indirect
```
stefanprodan, line 28: This is not Ok, 110 breaks backwards compatibly

stefanprodan (changes requested):

api/go.mod
makkes: Hu, I couldn't find an obvious commit in https://github.com/kubernetes/klog/compare/v2.100.1...v2.110.1 that would do that.

hiddeco: It appears to not break backwards compatibility when updated in conjunction with logr to v1.3.0. As my helm-controller build (which contains this bump) is running just fine.

stefanprodan: We could force controller-runtime to the new version if we bump logr also in okg/runtime, the version we are using is on klog 100 and logr 1.2 https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/go.mod

hiddeco: I can look into this in combination with a weird issue I observed earlier when `log-level` is set to `trace`:

```
{"level":"Level(-2)","ts":"2023-12-11T10:45:58.333Z","msg":"getting last revision of \"podinfo\"","controller":"helmrelease","controllerGroup":"helm.toolkit.fluxcd.io","controllerKind":"HelmRelease","HelmRelease":{"name":"podinfo","namespace":"podinfo"},"namespace":"podinfo","name":"podinfo","reconcileID":"daa23d2f-7b8a-4428-a61e-a816fe7d2846"}
```

makkes: Can someone please clarify on the backwards-incompatibility we're talking about here? The code compiles and tests fine so is it about the actual output it produces that's different with the new version?

hiddeco: "Backwards compatibility being" that klog v2.110 is only compatible with logr >=v1.3.0

stefanprodan: The backwards-incompatibility is about klog - logr, controller-runtime comes with logr 1.2 which doesn't work with klog 110. As I said above, we could force logr in pkg/runtime.

stefanprodan (approved):

PR #519, opened 2024-03-12
updating controller-gen to v0.14.0

souleb:
This also update the binary path to $GOBIN if set.


stefanprodan (approved):

config/crd/bases/image.toolkit.fluxcd.io_imagepolicies.yaml
```diff
@@ -1,9 +1,11 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.12.0
+    controller-gen.kubebuilder.io/version: v0.7.0
```
stefanprodan, line 7: This should be 0.14

stefanprodan (approved):

PR #557, opened 2024-05-02
docs: Remove references to aad pod identity

dipti-pai:
Remove references to the deprecated Azure AAD pod identity


stefanprodan (approved):
LGTM

PR #656, opened 2024-10-28
Doc: Image Repository: GCP: Direct resource access with WIF

JordanP:
GCP now favors direct resource access with WorkloadIdentity Federation, as opposed to service account impersonation.


JordanP:
If this is accepted, I'll update the[ other places](https://github.com/search?q=org%3Afluxcd%20%22The%20Artifact%20Registry%20service%20uses%20the%20permission%22&type=code) that need to be updated to match.

Issue #748, opened 2025-03-28
Image reflector controller uses a lot of ephemeral storage.


ilya-korol:
Hi, I'm using the Image Reflector Controller, which uses BadgerDB for storage. Over time, the controller creates a large number of 0000*.vlog files in the /data/ volume. Eventually, the storage gets full, leading to Kubernetes evicting the pod.

Tried setting --storage-value-log-file-size=2gb to reduce the number of vlog files and there was no effect.
I can see that many vlog files contain the same key-value pairs with minimal differences, because they all start with the same image name and its tags. It’s worth mentioning that we have a lot of images in ecr. 
There is no available configuration options to restrict either the total number of files or the maximum storage size used by Badger.

Is there a way to configure automatic cleanup or GC to prevent unbounded storage growth?
Are there any recommended best practices for managing storage in this use case?

Image Reflector Controller version: v0.31.2
Would appreciate any insights on how to resolve this issue. Thanks!

stealthybox:
[CHANGELOG.md](https://github.com/fluxcd/image-reflector-controller/blob/main/CHANGELOG.md) shows the effective changes between v0.31.2 and v0.34.0, and these are the commits since then: [v0.34.0..rc-c4913021](https://github.com/fluxcd/image-reflector-controller/commits/main/?since=2025-02-13&until=2025-05-23)

If you're able to upgrade in an affected cluster, could try using this RC image for the Garbage Collector?
- `ghcr.io/fluxcd/image-reflector-controller:rc-c4913021`
- `docker.io/fluxcd/image-reflector-controller:rc-c4913021`

You may also want to set `--log-level=debug` to see the badger GC logs

ilya-korol:
Hi, thank you for the changes. I will try to use the new image.

Number of image repos: ~500
Length and size of image tag lists: it's quite difficult to count, there are small repos (0-100), medium (100-2000) and large ones (2000+) (the largest is 10000k), average ~ 330 images per repos but it is a rough estimate.
CPU and Memory allocated: CPU requests - 100, limits 2000, Memory - requests 64Mb, limits 2Gb for reflector controller. Actual Memory Usage: 350Mb, with an increase to 950Mb every 40mins. 
Number of vlog files: 53 files (253Mb each)
Size of disk: 15Gb
Time period over which your disk-space runs out: 7-10 days

stefanprodan:
[...]

PR #757, opened 2025-04-30
Implement BadgerDB garbage collector with channels

stealthybox:
Garbage collection of the BadgerDB WAL/value-logs is our responsibility to orchestrate.
This patch implements it in a similar manner to dgraph.

`db.RunValueLogGC()` will rewrite only a single file.
We don't have to guard against concurrent runs of it, but we do need to run it multiple times until there is nothing left to write.

`gc.discardValueLogFiles()` does this, and we wrap it in a goroutine that waits `gc.Interval` between each run.
The use of a timer instead of a ticker here is so that the GC run can never overrun itself even if there are lots of files or the disk is very very very slow.

I've bounded the number of discards to prevent bugs in library-code and infrastructure we don't control from causing an infinite loop.

A stop channel is used to let `main()` wait on any running GC's.
I've observed from the logger that this lifecycle works.
There are no tests, but we could at minimum create a tests that creates, runs, and stops the garbage collector?

Actually testing the GC behavior would require writing a test where the garbage collector discards several value logs.
I haven't done this, but it feels potentially doable if we constrain the available memory to a badgerDB instance.

I'm unsure if I like this implemented better with a channel or a mutex -- what do you think?

Related #748


internal/database/badger_gc.go
```diff
@@ -0,0 +1,99 @@
...
+	// flow control
+	timer *time.Timer
+	stop  chan struct{}
+}
+
+// NewBadgerGarbageCollector creates and returns a new
+func NewBadgerGarbageCollector(db *badger.DB, interval time.Duration, log *logr.Logger) *BadgerGarbageCollector {
+	return &BadgerGarbageCollector{
+		DiscardRatio: 0.5, // must be a float between 0.0 and 1.0, inclusive
+		Interval:     interval,
```
stealthybox, line 42: Should `DiscardRatio` and `Interval` be configurable via controller flags?
We would need to validate values.

internal/database/badger_gc.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// To avoid GC Errors, call Stop() before closing the database.
+func (gc *BadgerGarbageCollector) Stop() {
+	gc.log.Info("Sending stop to Badger GC")
+	gc.stop <- struct{}{}
+	<-gc.stop
+}
+
+// upper bound for loop
+const maxDiscards = 1000
```
stealthybox, line 83: I don't know if this is a reasonable value of discards, but it felt like the controller could sample, delete, and rewrite 1000 value log files relatively quickly in its own goroutine without impacting the controller-runtime too much.

internal/database/badger_gc.go
```diff
@@ -0,0 +1,99 @@
...
+	// external deps
+	db  *badger.DB
+	log *logr.Logger
+	// flow control
+	timer *time.Timer
+	stop  chan struct{}
+}
+
+// NewBadgerGarbageCollector creates and returns a new
+func NewBadgerGarbageCollector(db *badger.DB, interval time.Duration, log *logr.Logger) *BadgerGarbageCollector {
```
stefanprodan, line 39: I would pass here the `ctrl.SetupSignalHandler()` and trigger the `Stop` function on `SIGTERM`.

stealthybox: resolved in redesign

main.go
```diff
@@ -132,7 +133,9 @@ func main() {
 		os.Exit(1)
 	}
 	defer badgerDB.Close()
+
 	db := database.NewBadgerDatabase(badgerDB)
+	badgerGC := database.NewBadgerGarbageCollector("badger-gc", badgerDB, 1*time.Minute, 0.7)
```
stealthybox, line 138: Should `DiscardRatio` and `Interval` be configurable via controller flags?
We would need to validate values.

stefanprodan: I think we should add a flag for the gc interval (type int representing minutes, default value `1`). Users should have the option to disable GC be setting the flag to `0`.

internal/database/badger_gc.go
```diff
@@ -0,0 +1,93 @@
...
+			// there is no more garbage to discard
+			gc.log.Info("Ran Badger GC", "discarded_vlogs", c)
+			return
+		}
+		if err != nil {
+			gc.log.Error(err, "Badger GC Error", "discarded_vlogs", c)
+			return
+		}
+	}
+	gc.log.Info("Ran Badger GC for maximum discards", "discarded_vlogs", maxDiscards)
```
stefanprodan, line 92:
```suggestion
	gc.log.V(1).Info("Ran Badger GC for maximum discards", "discarded_vlogs", maxDiscards)
```

Let's switch this to debug logging

stealthybox: specifically for the max discard edge case -- feels like it should be more of a visible warning (or even Error?)
if the GC is discarding 1000 files, that feels to me like it would be very behind, and the disk must be quite large?

Agreed for the normal discard case on Line 84 :+1:

internal/database/badger_gc_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/badger/v3"
+	"github.com/go-logr/logr"
+	"github.com/go-logr/logr/testr"
+)
+
+func TestBadgerGarbageCollectorDoesStop(t *testing.T) {
```
stealthybox, line 29: Example Successful run (2.3 seconds)
```
go test -run '^TestBadgerGarbageCollectorDoesStop$' github.com/fluxcd/image-reflector-controller/internal/database -v -count 1

=== RUN   TestBadgerGarbageCollectorDoesStop
badger 2025/05/07 00:04:06 INFO: All 0 tables opened in 0s
badger 2025/05/07 00:04:06 INFO: Discard stats nextEmptySlot: 0
badger 2025/05/07 00:04:06 INFO: Set nextTxnTs to 0
    badger_gc.go:61: test-badger-gc: "level"=0 "msg"="Starting Badger GC"
    badger_gc.go:84: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc_test.go:47: wrote tags successfully
    badger_gc.go:84: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc.go:84: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc_test.go:52: waiting for GC stop
    badger_gc.go:70: test-badger-gc: "level"=0 "msg"="Stopped Badger GC"
    badger_gc_test.go:57: GC Stopped
badger 2025/05/07 00:04:08 INFO: Lifetime L0 stalled for: 0s
badger 2025/05/07 00:04:08 INFO: 
Level 0 [ ]: NumTables: 01. Size: 216 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
--- PASS: TestBadgerGarbageCollectorDoesStop (2.06s)
PASS
ok  	github.com/fluxcd/image-reflector-controller/internal/database	2.355s
```

Synthetic Failure (sleep for a minute when stopped) (7.3 seconds)
```
go test -run '^TestBadgerGarbageCollectorDoesStop$' github.com/fluxcd/image-reflector-controller/internal/database -v -count 1

=== RUN   TestBadgerGarbageCollectorDoesStop
badger 2025/05/06 23:46:31 INFO: All 0 tables opened in 0s
badger 2025/05/06 23:46:31 INFO: Discard stats nextEmptySlot: 0
badger 2025/05/06 23:46:31 INFO: Set nextTxnTs to 0
    badger_gc.go:61: test-badger-gc: "level"=0 "msg"="Starting Badger GC"
    badger_gc.go:85: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc_test.go:47: wrote tags successfully
    badger_gc.go:85: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc.go:85: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc.go:85: test-badger-gc: "level"=0 "msg"="Ran Badger GC" "discarded_vlogs"=0
    badger_gc_test.go:52: waiting for GC stop
    badger_gc_test.go:55: GC did not stop
badger 2025/05/06 23:46:38 INFO: Lifetime L0 stalled for: 0s
badger 2025/05/06 23:46:38 INFO: 
Level 0 [ ]: NumTables: 01. Size: 216 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB
Level 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 4 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 5 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level 6 [B]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB
Level Done
--- FAIL: TestBadgerGarbageCollectorDoesStop (7.04s)
FAIL
FAIL	github.com/fluxcd/image-reflector-controller/internal/database	7.332s
FAIL
```

internal/database/badger_gc.go
stealthybox: Upped verbosity for the normal case -- lmk if you still want higher verbosity or an Error with Warning text for the Max Discards

stefanprodan: Ah I totally miss that this log happens only when max discard is hit. Yes let's change it to type error and warning text.

main.go
stealthybox: ✅

internal/database/badger_gc.go
stealthybox: ✅

stealthybox:
Notably, because we are using the controller manager, the Garbage Collector Runnable will only run when the controller is the Leader:
https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/manager/runnable_group.go#L55-L76

This is probably okay behavior, but if we want it to run the GC all the time (ex: when the controller becomes a follower), we can implement [`LeaderElectionRunnable`](https://github.com/kubernetes-sigs/controller-runtime/blob/b6c5897febe566008678f765ec5a3a1ea04a123a/pkg/manager/manager.go#L310-L315) and return `false` which will add it to the "Others" runnable group.

main.go
```diff
@@ -94,6 +99,7 @@ func main() {
 	flag.StringVar(&healthAddr, "health-addr", ":9440", "The address the health endpoint binds to.")
 	flag.StringVar(&storagePath, "storage-path", "/data", "Where to store the persistent database of image metadata")
 	flag.Int64Var(&storageValueLogFileSize, "storage-value-log-file-size", 1<<28, "Set the database's memory mapped value log file size in bytes. Effective memory usage is about two times this size.")
+	flag.Uint16Var(&gcInterval, "gc-interval", 1, "The number of minutes to wait between garbage collections. 0 disables the garbage collector.")
```
stefanprodan, line 102: I'm wandering if the 1m interval is too aggressive and if it will increase the CPU usage unnecessarily for most users. How about we go with 5m as the default, and we document [here](https://fluxcd.io/flux/components/image/options/) the flag (those that have a massive amount of images could set it to 1m).

stealthybox: After running some load-tests and considering how much smaller image tag data is -- I don't think it's very common for this GC code to run with the default BadgerDB table/LSM/vlog settings. The database has to contain a large number of image tags and use more than the default memory limit of 1GB to start flushing to vlog files.

I've updated the default interval to 10 minutes.
My hypothesis right now is end-users who are seeing this in production either have:
a) a very large number of image repos, or
b) a constantly changing set of repos with taglists larger than 1MB each

stefanprodan:
@stealthybox can you please rebase with upstream main

stealthybox:
rebased -- load-testing code/notes in #763 

It's a separate issue, but I think we could implement some much more conservative defaults right now for badger.
I've only been running synthetic tests with our controller libraries, but I think it could be pretty easy to OOM kill the controller with a small number of repos over time.

stefanprodan (approved):
LGTM

Thanks @stealthybox

stefanprodan:
@stealthybox please rebase your fork with upstream main, I can't merge as your branch is out-of-date.

stealthybox:
something is broken on GitHub's side with this PR.
I've pushed 8ec426fca25ec8e42a5c6045e368e942c9d38110 to [`stealthybox/gc-channel`](https://github.com/stealthybox/image-reflector-controller/tree/gc-channel)
but this PR is still stuck on d01d01866e8f6275d1a9eeb682506d696b4ce009

stealthybox:
editing the PR and re-selecting `main` as the base branch fixed it:
https://stackoverflow.com/a/76590821

PR #758, opened 2025-04-30
Implement BadgerDB garbage collector with mutex

stealthybox:
Functionally, this is similar to #757 but implemented with a Mutex instead.

See the diff from the second commit in this PR.

Perhaps this is a more legible way of handling the GC state.
It does also give us some other options for yeilding the GC to the controller's signal interrupt.

Related #748

### DIFF
--- a/internal/database/badger_gc.go
+++ b/internal/database/badger_gc.go
@@ -17,6 +17,7 @@ package database
 
 import (
 	"errors"
+	"sync"
 	"time"
 
 	"github.com/dgraph-io/badger/v3"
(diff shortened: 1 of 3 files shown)


internal/database/badger_gc.go
```diff
@@ -0,0 +1,93 @@
...
+// runs.
+//
+// This is a non-blocking operation.
+// To stop the garbage collector, call Stop().
+func (gc *BadgerGarbageCollector) Start() {
+	gc.log.Info("Starting Badger GC")
+	gc.timer = time.AfterFunc(gc.Interval, func() {
+		gc.running.Lock()
+		gc.discardValueLogFiles()
+		gc.running.Unlock()
```
stealthybox, line 60: Alternatively, what if we only locked within the loop-body of `discardValueLogFiles()`?
Then the GC would stop work for each discard, and the controller could exit sooner -- it's an edge case and would require a little bit more sync with Stop().

stealthybox:
Closing this.
Stefan suggested the GC should shut down more proactively on SIGTERM instead of waiting for all of the controller manager to shut down first like we do when closing the database.
I realized then that I could just add the GC to the controller manager by implementing the Runnable interface.

Carrying forward in #757

PR #774, opened 2025-06-01
Fix ImagePolicy reconciler getting triggered when ImageRepository is not ready

### LINKED ISSUE #773: ImagePolicy errors after v0.35.0 upgrade
We upgraded to Flux v2.6.0 and Image Reflector Controller v0.35.0 yesterday, and most/all of our ImagePolicies have been firing `FluxImagePolicyDependencynotready` errors since. The controller logs indicate the cause of the alerts is "referenced ImageRepository has not been scanned yet", however the ImageRepositories themselves show no errors, and appear to be being scanned normally (as proof of that, there were several successful image updates during this period).

For reference, our ImageRepositories are AWS ECR, and we're authing with IRSA (i.e. an AWS IAM role annotation applied to the Image Reflector Controller's SA).

For the moment we've reverted Image Reflector Controller to v0.34.0 and that has fixed the issue.

matheuscscp:
Fixes #773

### DIFF
--- a/internal/controller/imagepolicy_controller.go
+++ b/internal/controller/imagepolicy_controller.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #776, opened 2025-06-01
Fix ImagePolicy reconciler spamming `no tags in database` after a restart

matheuscscp:
This PR adds drift detection to the ImageRepository reconciler when the database is empty. After a controller restart (when the `/data` dir is not stored in a persistent volume), the policies will emit at most one `no tags in database` error event, then wait for the ImageRepository reconciler to fill the database.

### DIFF
--- a/internal/controller/imagepolicy_controller.go
+++ b/internal/controller/imagepolicy_controller.go
@@ -330,16 +330,6 @@ func (r *ImagePolicyReconciler) reconcile(ctx context.Context, sp *patch.SerialP
 		return
 	}
 
-	// Proceed only if the ImageRepository has scan result.
-	if !conditions.IsReady(repo) {
-		// Mark not ready but don't requeue. When the repository becomes ready,
-		// it'll trigger a policy reconciliation. No runtime error to prevent
-		// requeue.
-		conditions.MarkFalse(obj, meta.ReadyCondition, imagev1.DependencyNotReadyReason, "referenced ImageRepository has not been scanned yet")
-		result, retErr = ctrl.Result{}, nil
-		return
-	}
-
 	// Check if the image is valid and mark stalled if not.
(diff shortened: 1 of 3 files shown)


internal/controller/imagepolicy_controller.go
```diff
@@ -371,7 +361,7 @@ func (r *ImagePolicyReconciler) reconcile(ctx context.Context, sp *patch.SerialP
 		// If there's no tag in the database, mark not ready and retry.
 		if err == errNoTagsInDatabase {
 			conditions.MarkFalse(obj, meta.ReadyCondition, imagev1.DependencyNotReadyReason, "%s", err)
-			result, retErr = ctrl.Result{}, err
+			result, retErr = ctrl.Result{}, nil
```
stefanprodan, line 364: We no longer retry, please update the comment above

internal/controller/imagerepository_controller.go
```diff
@@ -290,7 +290,8 @@ func (r *ImageRepositoryReconciler) reconcile(ctx context.Context, sp *patch.Ser
 	// Scan the repository if it's scan time. No scan is a no-op reconciliation.
 	// The next scan time is not reset in case of no-op reconciliation.
 	if ok {
-		reconcile.ProgressiveStatus(false, obj, meta.ProgressingReason, "scanning: %s", reasonMsg)
+		drift := reasonMsg == scanReasonEmptyDatabase
```
stefanprodan, line 293: Add a comment here please explaining that this sets the Ready condition to Unknown when the db is empty

internal/controller/imagepolicy_controller.go
matheuscscp: Fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #780, opened 2025-06-01
Store checksum of ImageRepository tags and trigger ImagePolicy watch only when it changes

### LINKED ISSUE #778: Add `.status.revision` field to ImageRepository
To track changes in upstream tags, the `ImageRepository` reconciler should compute a checksum from the array of tags before updating the database. The checksum should be persisted in the `.status.revision` field of the `ImageRepository`.  The  `ImageRepository.status.revision` will be used by the `ImagePolicy` watcher to reliably determine when it should query the database to evaluate the policies.

matheuscscp:
Closes #778 

There's already a field called `.spec.lastScanResult` with other information about the scan that is pretty close to `revision` (e.g. `tagCount`, `latestTags`), so I figured it would make more sense to put the revision inside it as well.

### DIFF
--- a/api/v1beta2/imagerepository_types.go
+++ b/api/v1beta2/imagerepository_types.go
@@ -28,6 +28,7 @@ import (
 const ImageRepositoryKind = "ImageRepository"
 
 // Deprecated: Use ImageFinalizer.
+// TODO: Remove in v1.
 const ImageRepositoryFinalizer = ImageFinalizer
 
 // ImageRepositorySpec defines the parameters for scanning an image
@@ -115,10 +116,26 @@ type ImageRepositorySpec struct {
 	Insecure bool `json:"insecure,omitempty"`
 }
 
+// ScanResult contains information about the last scan of the image repository.
+// TODO: Make all fields except for LatestTags required in v1.
 type ScanResult struct {
-	TagCount   int         `json:"tagCount"`
-	ScanTime   metav1.Time `json:"scanTime,omitempty"`
-	LatestTags []string    `json:"latestTags,omitempty"`
+	// Revision is a stable hash of the scanned tags.
+	// +optional
+	Revision string `json:"revision"`
+
+	// TagCount is the number of tags found in the last scan.
+	// +required
+	TagCount int `json:"tagCount"`
+
+	// ScanTime is the time when the last scan was performed.
+	// +optional
+	ScanTime metav1.Time `json:"scanTime"`
+
+	// LatestTags is a small sample of the tags found in the last scan.
+	// It's the first 10 tags when sorting all the tags in descending
+	// alphabetical order.
+	// +optional
+	LatestTags []string `json:"latestTags,omitempty"`
 }
 
 // ImageRepositoryStatus defines the observed state of ImageRepository
--- a/config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
+++ b/config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
@@ -483,13 +483,23 @@ spec:
                 description: LastScanResult contains the number of fetched tags.
                 properties:
                   latestTags:
+                    description: |-
+                      LatestTags is a small sample of the tags found in the last scan.
+                      It's the first 10 tags when sorting all the tags in descending
+                      alphabetical order.
                     items:
                       type: string
                     type: array
+                  revision:
+                    description: Revision is a stable hash of the scanned tags.
+                    type: string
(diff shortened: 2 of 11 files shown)


api/v1beta2/imagerepository_types.go
```diff
@@ -116,6 +116,7 @@ type ImageRepositorySpec struct {
 }
 
 type ScanResult struct {
+	Revision   string      `json:"revision"`
```
stefanprodan, line 119: This should be optional in v1beta2. When we promote the API to v1 we’ll make it required. Please add a comment about v1 in here so we’ll not forget about it.

matheuscscp: Fixed, thanks!

api/v1beta2/imagerepository_types.go
```diff
@@ -116,6 +116,11 @@ type ImageRepositorySpec struct {
...
 
 type ScanResult struct {
+	// Revision is a stable hash of the scanned tags.
+	// TODO: Remove optional in API v1.
+	// +optional
+	Revision string `json:"revision"`
+
 	TagCount   int         `json:"tagCount"`
 	ScanTime   metav1.Time `json:"scanTime,omitempty"`
 	LatestTags []string    `json:"latestTags,omitempty"`
```
stefanprodan, line 126: Can you please add comments to each field and set `+optional`.  Missed this when we introduced  `ScanResult`. Also please add a comment to the `ScanResult` type.

matheuscscp: The only place `.status.lastScanResult` receives a value, it's here:

```go
	scanTime := metav1.Now()
	obj.Status.LastScanResult = &imagev1.ScanResult{
		Revision:   checksum,
		TagCount:   len(filteredTags),
		ScanTime:   scanTime,
		LatestTags: latestTags,
	}
```

Are these fields really optional? They're always set :thinking: 

`.status.lastScanResult` is already optional itself:

```go
	// LastScanResult contains the number of fetched tags.
	// +optional
	LastScanResult *ScanResult `json:"lastScanResult,omitempty"`
```

stefanprodan: I guess latestTags can be empty, what if the repo has no images? But all others should be required, we'll change them in v1.

config/crd/bases/image.toolkit.fluxcd.io_imagerepositories.yaml
```diff
@@ -483,16 +483,24 @@ spec:
...
+                    description: ScanTime is the time when the last scan was performed.
                     format: date-time
                     type: string
                   tagCount:
+                    description: TagCount is the number of tags found in the last
+                      scan.
                     type: integer
-                required:
-                - tagCount
                 type: object
```
matheuscscp, line 504: Hmm it's probably better if we don't add `+optional` to this field now... I'll revert the change only for this one,

api/v1beta2/imagerepository_types.go
```diff
@@ -115,10 +116,26 @@ type ImageRepositorySpec struct {
...
+	// +optional
+	Revision string `json:"revision"`
+
+	// TagCount is the number of tags found in the last scan.
+	// +required
+	TagCount int `json:"tagCount"`
+
+	// ScanTime is the time when the last scan was performed.
+	// +optional
+	ScanTime metav1.Time `json:"scanTime"`
```
matheuscscp, line 132: Since we're adding `+optional` to this field, I'm removing `,omitempty` because that has no effect for struct fields in `json` serialization (only for struct pointers), and `+optional` keeps the field being optional in the CRD (TIL: `,omitempty` makes fields optional in the CRD!).

internal/controller/imagerepository_controller_test.go
```diff
@@ -277,66 +279,74 @@ func TestImageRepositoryReconciler_scan(t *testing.T) {
...
+		db            *mockDatabase
+		proxyURL      *url.URL
+		wantErr       string
+		wantChecksum  string
+		wantTags      []string
 	}{
 		{
 			name:    "no tags",
 			wantErr: "404 Not Found",
 		},
```
matheuscscp, line 295: @stefanprodan There's already a test for empty tags. The registry returns 404 (at least the one used in the tests). I tried adding a test like this and the behavior is the same:

```go
		{
			name:     "empty tags",
			tags:     []string{},
			db:       &mockDatabase{},
			wantTags: []string{},
		},
```

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #786, opened 2025-06-20
Remove deprecated auto-login flags

cappyzawa:
## Summary

Remove support for deprecated auto-login controller flags as part of preparing image automation APIs for GA.

This addresses task 3 "Remove deprecated code paths (e.g. auto-login flags in IRC)" from https://github.com/fluxcd/flux2/issues/5411.

Removes the following flags that have been deprecated since v0.25.0:
- `--aws-autologin-for-ecr`
- `--gcp-autologin-for-gcr` 
- `--azure-autologin-for-acr`

## Changes

- Remove flag definitions and variables from main.go
- Remove `DeprecatedLoginOpts` field from `ImageRepositoryReconciler`
- Remove deprecated login options handling from registry options
- Clean up unused imports (`errors`, `aws`, `azure`, `gcp` packages)
- Clarify switch statement logic for provider auto-login

## Breaking Change

This is a **planned breaking change** for flags that have been deprecated since v0.25.0 (February 2023). In v0.26.0, it was announced that these flags would be removed "at least one minor version after this release" - sufficient time has now passed (v0.35.x).

The flags currently show error messages when used, and users have been advised to migrate to the `.spec.provider` field.

After this change, using these flags will cause the controller to fail with "flag provided but not defined" error.

## Migration

Users must update their ImageRepository objects to use the `.spec.provider` field instead:

```yaml
kind: ImageRepository
metadata:
  name: my-app
spec:
  image: 123456789012.dkr.ecr.us-east-1.amazonaws.com/my-app
  provider: aws  # Use this instead of --aws-autologin-for-ecr flag
```

Related to https://github.com/fluxcd/flux2/issues/5411 (task 3)

### DIFF
--- a/internal/registry/options.go
+++ b/internal/registry/options.go
@@ -31,9 +31,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluxcd/pkg/auth"
-	"github.com/fluxcd/pkg/auth/aws"
-	"github.com/fluxcd/pkg/auth/azure"
-	"github.com/fluxcd/pkg/auth/gcp"
 	authutils "github.com/fluxcd/pkg/auth/utils"
 	"github.com/fluxcd/pkg/cache"
 
@@ -57,7 +54,7 @@ type AuthOptionsGetter struct {
 }
 
 func (r *AuthOptionsGetter) GetOptions(ctx context.Context, repo *imagev1.ImageRepository,
-	involvedObject *cache.InvolvedObject, deprecatedLoginOpts ...auth.Provider) ([]remote.Option, error) {
+	involvedObject *cache.InvolvedObject) ([]remote.Option, error) {
 	timeout := repo.GetTimeout()
 	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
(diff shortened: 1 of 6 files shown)


internal/registry/options.go
```diff
@@ -104,9 +104,10 @@ func (r *AuthOptionsGetter) GetOptions(ctx context.Context, repo *imagev1.ImageR
 		if proxyURL != nil {
 			opts = append(opts, auth.WithProxyURL(*proxyURL))
 		}
+		// Auto-login is only supported for specific cloud providers
 		switch provider := repo.GetProvider(); provider {
```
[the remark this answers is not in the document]
cappyzawa: Thanks for your review.

Fixed via https://github.com/fluxcd/image-reflector-controller/pull/786/commits/8a8add8e17a0bbf7dde32c280e442e0f86622be6 .

internal/registry/options.go
```diff
@@ -98,36 +97,23 @@ func (r *AuthOptionsGetter) GetOptions(ctx context.Context, repo *imagev1.ImageR
 			return nil, err
 		}
 		authenticator, authErr = secret.AuthFromSecret(authSecret, ref)
-	} else {
+	} else if provider != "generic" {
```
matheuscscp, line 100:
```suggestion
	} else if provider := repo.GetProvider(); provider != "generic" {
```

Can declare the variable like this inside the else block since it's not used anywhere else

internal/registry/options.go
```diff
@@ -98,36 +97,23 @@ func (r *AuthOptionsGetter) GetOptions(ctx context.Context, repo *imagev1.ImageR
...
+			serviceAccount := client.ObjectKey{
+				Name:      repo.Spec.ServiceAccountName,
+				Namespace: repo.GetNamespace(),
 			}
+			opts = append(opts, auth.WithServiceAccount(serviceAccount, r.Client))
+		}
+		if r.TokenCache != nil {
+			opts = append(opts, auth.WithCache(*r.TokenCache, *involvedObject))
 		}
+		authenticator, authErr = authutils.GetArtifactRegistryCredentials(ctx, provider, repo.Spec.Image, opts...)
```
matheuscscp, line 116:
```suggestion
		authenticator, authErr = authutils.GetArtifactRegistryCredentials(ctx, provider, repo.Spec.Image, opts...)
		if authenticator == nil && authErr == nil {
			return nil, fmt.Errorf("unsupported provider: %s", provider)
		}
```

internal/registry/options.go
cappyzawa: Fixed via https://github.com/fluxcd/image-reflector-controller/pull/786/commits/a2b4049945aade03c5216671a8cf65d37ed5cdc6 .

internal/registry/options.go
cappyzawa: Sorry for not fully understanding your intent initially!

Fixed via a2b4049e17a0bbf7dde32c280e442e0f86622be6:
- Moved provider variable declaration into the if statement 
- Added error check for unsupported provider after GetArtifactRegistryCredentials

Since this introduces a new error case for unsupported providers, I also added a test case in 5ffd506e17a0bbf7dde32c280e442e0f86622be6 to ensure the error is properly handled.

cappyzawa:
That's a great point about improving the error handling in `GetArtifactRegistryCredentials()`. 

Should we:
1. Keep the current `if authenticator == nil && authErr == nil` check for now, merge this PR, and then handle the pkg improvement + cleanup in a follow-up?
2. Or remove this check now and wait for the fluxcd/pkg changes before merging this PR?

I'm happy to work on the fluxcd/pkg changes either way, just want to clarify the preferred approach for this PR.

cappyzawa:
I've created https://github.com/fluxcd/pkg/pull/946 to handle the error handling improvement in `GetArtifactRegistryCredentials()`.

matheuscscp:
@cappyzawa I can run `go get github.com/fluxcd/pkg/auth` and `go mod tidy`, should upgrade to `v0.18.0`. Then you can remove the `if` we discussed and squash, then we're good to merge :ok_hand:

matheuscscp:
Actually, these commit messages and content look good :)

matheuscscp:
@cappyzawa Two more tiny PRs and we're done with the `pkg/auth` improvement.

## flux2 (CLI)

Bump to `auth` `v0.18.0`, remove this [check](https://github.com/fluxcd/flux2/blob/8ae0aaa46c43e29b6b1ba6c1e205fc13c0a6e4ad/cmd/flux/oci.go#L41-L43) and replace this [block](https://github.com/fluxcd/flux2/blob/8ae0aaa46c43e29b6b1ba6c1e205fc13c0a6e4ad/cmd/flux/push_artifact.go#L232-L242) with a call to `loginWithProvider()` .

## source-controller

Bump to `auth` `v0.18.0`, and this [block](https://github.com/fluxcd/source-controller/blob/97c995b8c8e3a6c9c31cb4f1e8323428a03d07d3/internal/helm/getter/client_opts.go#L142-L144) can be simplified to just `hrOpts.Authenticator = authenticator`.

PR #798, opened 2025-07-23
Fix failing test case in options test

dgunzy:
`make test` was failing in the main branch - there is no error here with a valid provider during the test run. 

@matheuscscp I am working on `flux reconcile image policy` and wanted the tests to work out of the box 😄 
Let me know if I should fix it a different way.

### DIFF
--- a/internal/registry/options_test.go
+++ b/internal/registry/options_test.go
@@ -150,7 +150,7 @@ func TestNewAuthOptionsGetter_GetOptions(t *testing.T) {
 				Image:    "123456789000.dkr.ecr.us-east-2.amazonaws.com/test",
 				Provider: "aws",
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "cloud provider repo without login",


matheuscscp:
Hi @dgunzy :wave: 

This test is for contextual authentication. It's not passing for you locally because your machine is likely configured with AWS authentication, and the test is expecting the error of *not having* authentication configured in the context it runs on, which is the case for the CI.

You have a few alternatives:
* Ignore this test when running `make test` locally. It's doing what it should, so no need to worry.
* Undo your local AWS authentication configuration. That's probably inconvenient for you, so I'd not recommend it.
* You can file another PR in this repo contributing something else, then you'll be able to use the CI without my manual approval. You could for example file a PR for `Remove .status.latestImage from ImagePolicy status`, which is literally just remove things from the code, and associated tests and docs.

I'm not sure however why you need to work on this repo for `flux reconcile image policy`? Should be a file in the `fluxcd/flux2`

dgunzy:
Hi! Sorry - I did not think about the tests using my aws credentials -- I will close this 😅.. 

So i was clearly thinking about the `flux reconcile image policy` wrong - when I created a file similar to reconcile image policy, there needed to be some different functions added to fulfill the `reconcilable` interface.  Should all of that be done in the main flux codebase? 

eg adding 
```
func (a imagePolicyAdapter) setSuspended() {
	a.Spec.Suspend = true
}
``` 
Lints to : `a.Spec.Suspend undefined (type "github.com/fluxcd/image-reflector-controller/api/v1beta2".ImagePolicySpec has no field or method Suspend` 

I don't think I am understanding the right approach - would you mind quickly outlining the right way to do this? 

Thanks for the patience @matheuscscp !!!

PR #806, opened 2025-08-14
Remove TLS ServerName pinning in TLS config creation

cappyzawa:
ref: https://github.com/fluxcd/pkg/pull/1007, https://github.com/fluxcd/pkg/pull/993


matheuscscp (approved):
LGTM! 🚀

Issue #811, opened 2025-08-31
Upgrade Badger to v4


stefanprodan:
We should upgrade from `github.com/dgraph-io/badger/v3` to `github.com/dgraph-io/badger/v4` and figure out if there are any changes needed to https://github.com/fluxcd/image-reflector-controller/tree/main/internal/database

PR #763, opened 2025-05-23
Loadtest BadgerGC

stealthybox:
related #757

PR #816, opened 2025-09-10
Upgrade to Badger v4

stealthybox:
Commits:
- **Handle controller-runtime failing to add BadgerGC**
- **Upgrade badger to v4.8.0**

badger v3.x -> v4.0 isn't a breaking change,
just an update from calver to semver
(no binary format change)
https://github.com/hypermodeinc/badger/blob/main/CHANGELOG.md#400---2023-02-27

The v4.x series does contain 2 breaking changes:

1. **change to metrics names + new metrics**
      badger uses the go expvar format.
      We enable these on our Badger by default, but we don't actually expose them in controller-runtime Prometheus format...
      no impact
2. **change to runtime errors deprecating pkg/errors**
      investigated this:  our dependency on their error types for the GC is covered in a test and it's still functioning + unchanged)

Loadtest also rebased and good: https://github.com/fluxcd/image-reflector-controller/pull/763#issuecomment-3273460920

Closes: https://github.com/fluxcd/image-reflector-controller/issues/811


stefanprodan (approved):
LGTM

Thanks @stealthybox

PR #824, opened 2025-09-19
Fix controllers not watching reconcile requests from annotations

lukas8219:
While working on https://github.com/fluxcd/flux2/pull/5492 we noticed that the ImagePolicy controller was lacking the predicate for watching reconcile requests from the annotation. This bug escaped due to a weakness on the controller test for this scenario where the object was being created with the annotation, so the first reconciliation was copying the annotation to the status field. After changing this test to create the object without the annotation and to add it later, the test started to fail, making the bug surface.

We are also improving the same test for the ImageRepository controller, and modifying its predicate setup to match the style of the majority of the other controllers.


matheuscscp:
@lukas8219 Thanks for catching this!

PR #880, opened 2026-04-13
fix: clear Result when Reconcile returns an error

### LINKED ISSUE #661: small reconciler bug (result + error returned) when error occurred
Hello,

It seems that the image-reflector-controllers reconciler is always returning the result and error object without checking whether the error is not nil.
I do see these messages upon error:
```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```
I suggest to change this:
https://github.com/fluxcd/image-reflector-controller/blob/a71ffb31af5dd74399b2b77929383ab9d03acee6/internal/controller/imagerepository_controller.go#L192

to something like this:
```go
if retErr != nil {
  return nil, retErr
}

return result, nil
```

Same can be applied for the imagepolicy controller: https://github.com/fluxcd/image-reflector-controller/blob/a71ffb31af5dd74399b2b77929383ab9d03acee6/internal/controller/imagepolicy_controller.go#L198

There might be other flux components doing the same, but I didn't check them.

alliasgher:
## Summary

Both `ImageRepositoryReconciler.Reconcile` and `ImagePolicyReconciler.Reconcile` wrap the sub-reconciler with a named-return form that always emits both values:

```go
result, retErr = r.reconcile(ctx, ...)
return
```

controller-runtime emits the warning referenced in #661 when a `Reconcile` call returns both a non-zero `Result` and a non-nil error — the `Result` is ignored in that case and the object is re-queued with the error's exponential backoff. Populating a `RequeueAfter` and then returning an error gives the misleading impression that the caller's requested interval applies.

Drop the `Result` when there is an error so controller-runtime sees `(ctrl.Result{}, retErr)` and the warning no longer fires. Behaviour after the change is what controller-runtime was doing anyway (the `Result` was being discarded); this just avoids the log noise.

Fixes #661

Signed-off-by: Ali <alliasgher123@gmail.com>

### DIFF
--- a/internal/controller/imagepolicy_controller.go
+++ b/internal/controller/imagepolicy_controller.go
@@ -280,6 +280,12 @@ func (r *ImagePolicyReconciler) reconcile(ctx context.Context, sp *patch.SerialP
 	defer func() {
 		rs := pkgreconcile.NewResultFinalizer(isSuccess, readyMsg)
 		retErr = rs.Finalize(obj, result, retErr)
+		// controller-runtime logs a warning and ignores Result when Reconcile
+		// returns both a non-zero Result and a non-nil error. Clear the Result
+		// whenever there is an error so the log stays clean.
+		if retErr != nil {
+			result = ctrl.Result{}
+		}
 
 		// Presence of reconciling means that the reconciliation didn't succeed.
 		// Set the Reconciling reason to ProgressingWithRetry to indicate a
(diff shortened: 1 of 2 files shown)


stefanprodan:
We should fix this inside the `reconcile` func, instead of adding a bandaid.

stefanprodan:
@alliasgher the commit should be:

```
fix: clear Result when Reconcile returns an error

Signed-off-by: Ali <alliasgher123@gmail.com>
Assisted-by: claude-sonnet-4-6
```

The fix should happen inside the `r.reconcile` functions, where in all places we return an error, we should do:

```go
return nil, err
```

alliasgher:
> @alliasgher the commit should be:
> 
> ```
> fix: clear Result when Reconcile returns an error
> 
> Signed-off-by: Ali <alliasgher123@gmail.com>
> Assisted-by: claude-sonnet-4-6
> ```
> 
> The fix should happen inside the `r.reconcile` functions, where in all places we return an error, we should do:
> 
> ```go
> return nil, err
> ```

@stefanprodan done!

PR #900, opened 2026-06-18
Fix timeout created more than once

matheuscscp:
After pulling https://github.com/fluxcd/pkg/pull/1244 in, this bug in the timeout has surfaced. We were creating two separate contexts with the `.spec.timeout`, which has two problems:

* The first context, for setting up the auth options, gets canceled before the creds are fetched. Since the `ctx` in `auth/utils.GetArtifactRegistryCredentials` is now wrapped in a closure that is called lazily, this cancellation bubbles up when the OCI client tries to `Authorize`. (`Authorize` does not receive a `ctx`, which is very unfortunate from `go-containerregistry`)
* The effective timeout is not the one promised by the `.spec.timeout` API field, but up to almost twice that value.

Now the timeout is applied to both auth setup + registry calls, as it should.

Note: This bug does not exist in source-controller.

### DIFF
--- a/internal/controller/imagepolicy_controller.go
+++ b/internal/controller/imagepolicy_controller.go
@@ -459,12 +459,19 @@ func (r *ImagePolicyReconciler) fetchDigest(ctx context.Context,
 		Namespace: obj.GetNamespace(),
 		Operation: cache.OperationReconcile,
 	}
+
+	// The timeout must span both building the auth options and the registry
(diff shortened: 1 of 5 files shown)


matheuscscp:
Tested in my EKS clusters, fix works:

Screenshot from 2026-06-18 13-57-01

stefanprodan (approved):
LGTM

Great find! thanks @matheuscscp

